PR #11, opened 2025-09-30
Added widget conditions

ivo-ritense:


ivo-ritense:
Recreating PR with cherry pick

PR #12, opened 2025-09-30
Added widget conditions

ivo-ritense:
https://github.com/generiekzaakafhandelcomponent/atlas-internal/issues/5

How to test:
- For a case widget tab, set up a widget that has a condition based on something in the document (e.g. voornaam == 'Peter')
- Create a case for that case definition.
- Set the property to match the required value.
- Check the widget tab that the widget is there.
- Set the property to another value that doesn't match.
- Check the widget tab that the widget isn't there.
- Configure a widget that has no conditions
- Check the widget tab that the widget is there.

Other tests:
- Run the application with rabbitMQ (default for development)
- Open the same case with conditional widgets in 2 browsers

Widget updates
- Set the property to a value that doesn't match.
- Set the property to another value that doesn't match.
- Check that the data is updated in the other browser but the whole widget collection isn't refreshed with the spinner.
- et the property to another value that does match.
- Check that the widget is displayed in both browsers.

Summary form
- create a summary form showing data
- set up the summary tab
- Update a case in 1 browser, see that the summary form was updated in both browsers

formio form
- create a formio form showing data
- set up a case tab using the form
- Update a case in 1 browser, see that the summary form was updated in both browsers

Possible regressions:
- Check the dashboard conditions to see if that still works
- Check the translations of the operators (==, !=, etc.) in the dashboard widget settings


backend/case/src/main/kotlin/com/ritense/case_/rest/dto/CaseWidgetTabDto.kt
```diff
@@ -41,41 +42,49 @@ data class CaseWidgetTabDto(
...
                     }
             )
         }
+
+        fun andCheck(
+            first: (CaseWidgetTabWidget, JsonSchemaDocument) -> Boolean,
+            second: (CaseWidgetTabWidget, JsonSchemaDocument) -> Boolean,
+        ): (CaseWidgetTabWidget, JsonSchemaDocument) -> Boolean {
+            return { widget, document -> first(widget, document) && second(widget, document) }
+        }
```
Klaas-Ritense: The use of `JsonSchemaDocument` here, will make it difficult for the generic widgets for IKO

ivo-ritense: I agree, but looking at the ofWithContext method it was already used, this is just a helper function. How did you solve this for IKO?

frontend/projects/valtimo/case/src/lib/components/case-detail/tab/widgets/widgets.component.ts
```diff
@@ -53,10 +57,44 @@ export class CaseDetailWidgetsComponent implements OnInit, OnDestroy {
...
+  private _previousWidgetConfiguration: CaseWidgetsRes | null = null;
+
+  public readonly widgetConfiguration$ = combineLatest([
+    this._documentId$,
+    this._tabKey$,
+    this._documentUpdates$,
+  ]).pipe(
+    switchMap(([documentId, tabKey, documentUpdatedEvent]) => {
+      return this.widgetsApiService.getWidgetTabConfiguration(documentId, tabKey).pipe(
+        map(configuration => {
```
mbritense: please abstract this to private method

frontend/projects/valtimo/case/src/lib/models/case-widget.model.ts
```diff
@@ -159,6 +160,21 @@ interface CustomCaseWidgetConfig {
   [componentKey: string]: Type<any>;
 }
 
+enum Operator {
+  NOT_EQUAL_TO = '!=',
+  EQUAL_TO = '==',
```
mbritense: Can these not be reused from dashboard? If not, move it to shared lib so it's only defined and used once

frontend/projects/valtimo/case/src/lib/models/case-widget.model.ts
```diff
@@ -159,6 +160,21 @@ interface CustomCaseWidgetConfig {
...
+  GREATER_THAN = '>',
+  GREATER_THAN_OR_EQUAL_TO = '>=',
+  LESS_THAN = '<',
+  LESS_THAN_OR_EQUAL_TO = '<=',
+}
+
+interface Condition {
```
mbritense: probably same as above

frontend/projects/valtimo/layout/src/lib/components/widget-container/widget-container.component.ts
```diff
@@ -73,13 +74,26 @@ export class WidgetContainerComponent implements AfterViewInit, OnDestroy {
...
+  public readonly reload$ = new BehaviorSubject<boolean>(false);
+  public readonly loaded$ = merge(this.reload$, this.widgetLayoutService.loaded$.pipe(delay(400)));
 
   private _observer!: ResizeObserver;
 
   constructor(private readonly widgetLayoutService: WidgetLayoutService) {}
 
-  public ngAfterViewInit(): void {
+  private resetLayout(): void {
+    if (this._observer) {
```
mbritense: preferable to do if (!this._observer) return

frontend/projects/valtimo/layout/src/lib/components/widget-formio/widget-formio.component.ts
```diff
@@ -44,10 +55,38 @@ export class WidgetFormioComponent {
...
+  public set refreshForm(value: EventEmitter<void> | null) {
+    if (value === this._refreshEmitter) {
+      return;
+    }
+
+    this._refreshSubscription?.unsubscribe();
+    this._refreshSubscription = null;
+    this._refreshEmitter = value;
+
+    if (!value) {
```
mbritense: no curly brackets needed, so code guidelines

frontend/projects/valtimo/layout/src/lib/components/widget-formio/widget-formio.component.ts
```diff
@@ -44,10 +55,38 @@ export class WidgetFormioComponent {
...
+  private readonly _refreshTrigger$ = new BehaviorSubject<void>(undefined);
+
+  private _refreshEmitter: EventEmitter<void> | null = null;
+  private _refreshSubscription: Subscription | null = null;
+  private _widgetUuid: string | null = null;
+  private _hasSignalledExternalDataReady = false;
+
+  @Input()
+  public set refreshForm(value: EventEmitter<void> | null) {
+    if (value === this._refreshEmitter) {
```
mbritense: no curly brackets needed, so code guidelines

frontend/projects/valtimo/case/src/lib/components/case-detail/tab/widgets/components/formio/case-widget-formio.component.ts
```diff
@@ -56,6 +64,8 @@ export class CaseWidgetFormioComponent extends WidgetProcess {
     return this._widgetConfigurationSubject$.pipe(filter(config => !!config));
   }
 
+  public refreshForm = new EventEmitter<void>();
```
teodor-ritense:
```suggestion
  public readonly refreshForm = new EventEmitter<void>();
```

frontend/projects/valtimo/layout/src/lib/components/widget-formio/widget-formio.component.ts
```diff
@@ -73,13 +113,21 @@ export class WidgetFormioComponent {
       ])
     ),
     tap(() => {
-      this.widgetLayoutService.setWidgetWithExternalDataReady(this.widgetUuid);
+      if (!this._widgetUuid) {
```
teodor-ritense: No curly brackets needed

backend/case/src/main/kotlin/com/ritense/case_/rest/dto/CaseWidgetTabDto.kt
Klaas-Ritense: `ofWithContext` is used for PBAC, which is only partially implemented for IKO. So `ofWithContext` is also a problem for IKO

Klaas-Ritense (approved):
BE looks good

backend/case/src/main/kotlin/com/ritense/case_/rest/dto/CaseWidgetTabDto.kt
ivo-ritense: I removed the method and did it in another way.

frontend/projects/valtimo/case/src/lib/components/case-detail/tab/widgets/widgets.component.ts
ivo-ritense: Done

frontend/projects/valtimo/case/src/lib/models/case-widget.model.ts
ivo-ritense: I moved these to shared and use them in both dashboard and widgets. Also moved all translations to the general translation files so they are re-used everywhere.

frontend/projects/valtimo/case/src/lib/models/case-widget.model.ts
ivo-ritense: There's a small difference in names here (`path` vs `queryPath` etc.) that prevents re-use here. I can't change this currently because that would mean breaking changes in the API for dashboard.

I did however move these to the shared library to that next time we need conditions for something we can more easily re-use.

mbritense:
FE looks good

sofiaIvarsRitense (approved):

Klaas-Ritense (approved):

PR #33, opened 2025-10-06
GH-332: Fix error: Verzoek plugin configuration auto-deployment order

Klaas-Ritense:


backend/zgw/verzoek/src/main/kotlin/com/ritense/verzoek/VerzoekPlugin.kt
```diff
@@ -78,8 +79,9 @@ class VerzoekPlugin(
                             caseDefinitionKey = property.caseDefinitionKey,
                             caseDefinitionVersionTag = property.caseDefinitionVersionTag,
                         )
-                        require(caseDefinitions.isNotEmpty()) {
-                            "No case definition found for '${property.caseDefinitionKey}:${property.caseDefinitionVersionTag}'."
+                        if (caseDefinitions.isEmpty()) {
```
marijnritense, line 82: Won't this just lead the documentDefinitions being empty, and therefore not lead to any validation of the json pointers? If this is the case, maybe we can retrigger the validateProperties method at a later stage.

Klaas-Ritense: Luckily, the `validateProperties` is re-triggered later, thanks to the `EventType.UPDATE`. Also the `validateProperties` is always done when working through the UI.

ivo-ritense (approved):

ivo-ritense (approved):

PR #34, opened 2025-10-06
Added method to update a document with db lock

ivo-ritense:
Recreation of PR from old backend repository: https://github.com/valtimo-platform/valtimo-backend-libraries/pull/1984

## Changes
----

### Core Implementation:

- Add `findByIdForUpdate()` with `@Lock(PESSIMISTIC_WRITE)` to JsonSchemaDocumentRepository
- Add `modifyDocumentAtomic()` method with configurable timeout to DocumentService
- Create `DocumentProperties` for Spring Boot configuration

### ValueResolver Integration:

- Update DocumentJsonValueResolverFactory to support both locking strategies
- Add retry mechanism (3 attempts, exponential backoff) for optimistic mode
- Enable per-operation locking configuration

### Configuration:

```
valtimo:
  document:
    locking:
      valueResolver:
        pessimisticEnabled: true
```

### Testing:
- Comprehensive unit tests for retry mechanisms
- Integration tests for concurrent update scenarios
- Performance validation

### Benefits

- Eliminates OptimisticLockingFailureException in workflow operations
- Maintains backward compatibility (optimistic by default)
- Configurable per environment and operation type

Resolves concurrent document update conflicts while preserving performance for low-contention scenarios.

https://github.com/valtimo-platform/valtimo-issues/issues/58


backend/case/src/main/java/com/ritense/document/config/DocumentProperties.java
```diff
@@ -0,0 +1,45 @@
...
+    public DocumentProperties(Boolean usePessimisticLocking, Integer lockTimeoutInMillis) {
+        this.usePessimisticLocking = usePessimisticLocking != null ? usePessimisticLocking : false;
+        this.lockTimeoutInMillis = lockTimeoutInMillis != null ? lockTimeoutInMillis : 30000;
+    }
+
+    public boolean isUsePessimisticLocking() {
+        return usePessimisticLocking;
+    }
+
+    public int getLockTimeoutInMillis() {
```
Klaas-Ritense: This property is never used

backend/case/src/main/java/com/ritense/document/config/DocumentProperties.java
```diff
@@ -0,0 +1,45 @@
...
+    private final boolean usePessimisticLocking;
+    private final int lockTimeoutInMillis;
+
+    @ConstructorBinding
+    public DocumentProperties(Boolean usePessimisticLocking, Integer lockTimeoutInMillis) {
+        this.usePessimisticLocking = usePessimisticLocking != null ? usePessimisticLocking : false;
+        this.lockTimeoutInMillis = lockTimeoutInMillis != null ? lockTimeoutInMillis : 30000;
+    }
+
+    public boolean isUsePessimisticLocking() {
```
Klaas-Ritense: This property is called `valtimo.document...`. People might assume that it is a global property which it is not. Since it only used in the value resolver and not in the `documentSerivce.modifyDocument(..)`.

Klaas-Ritense:
Problem, the pessimistic locking does not seem to work. I still get a stacktrace with an `ObjectOptimisticLockingFailureException` even though the value resolver is calling the new `documentService.updateDocumentAtomic(...)`
image

sdegroot:
> Problem, the pessimistic locking does not seem to work. I still get a stacktrace with an `ObjectOptimisticLockingFailureException` even though the value resolver is calling the new `documentService.updateDocumentAtomic(...)` image

@Klaas-Ritense would you be able to share your test set-up? I'm curious as to why it does not work

Klaas-Ritense:
During testing ([see screenshot](https://github.com/valtimo-platform/valtimo/pull/34#issuecomment-3375591333)), every concurrent update of the same document caused an OptimisticLockingException. This occurred because a cached instance of the Document was already present in the persistence context before calling `findByIdForUpdate(...)`. In that case, Hibernate reused the cached entity instead of executing the `SELECT ... FOR UPDATE` query, meaning the `PESSIMISTIC_WRITE` lock was never actually applied. As a result, the cached entity remained stale and updates failed with an optimistic locking error.

-- EDIT: changed to different solution after discussion with @ivo-ritense

ivo-ritense:
I still don't understand how this fails. If the transaction waits until it gets the row level lock on the database, it should get a certain version. No other transactions should be able to change that row until this transaction completes, right? So how would the version be changed in the mean time? 

Your fix seems to ignore an update that somehow get through where we don't expect it.

>any concurrent transaction that already changed the row increments the version and causes an optimistic-lock conflict when we try to save.

But there shouldn't be an concurrent transaction. Is this a hibernate caching problem or something, where even if we lock the row it doesn't refetch and instead gets it from the cache?

Klaas-Ritense (approved):
@ivo-ritense approves

PR #48, opened 2025-10-08
 Dockerize app modules and added pipelines to create test environments for each new PR

ThomasMinkeRitense:


backend/app/gzac/src/main/resources/config/application.yml
```diff
@@ -131,20 +143,14 @@ spring:
...
+        host: ${SPRING_RABBITMQ_HOST}
+        port: ${SPRING_RABBITMQ_PORT}
+        username: ${SPRING_RABBITMQ_USERNAME}
+        password: ${SPRING_RABBITMQ_PASSWORD}
         publisher-confirm-type: correlated
         publisher-returns: true
 
-logging:
-    level:
-        org.springframework.beans: DEBUG
```
ivo-ritense, line 143: Where do we set this info now? This logging is very useful for development.

backend/app/gzac/src/main/resources/config/application.yml
```diff
@@ -218,9 +211,6 @@ valtimo:
             - application/xml
             - image/jpeg
             - image/png
-    changelog:
-        pbac:
-            clear-tables: false
```
ivo-ritense, line 223: This is nice for development. Do we have to remove it?

ivo-ritense:
Can we add a development profile for the settings that should only apply when developing based on the app? Things like debug logging that we always want to enable.

backend/app/gzac/src/main/resources/config/application.yml
ThomasMinkeRitense: The default value is false, so this property didn't do anything.

backend/app/gzac/src/main/resources/config/application.yml
ThomasMinkeRitense: Added these properties to the .env.properties.

ThomasMinkeRitense:
> Can we add a development profile for the settings that should only apply when developing based on the app? Things like debug logging that we always want to enable.

The local properties go in .env.properties, that will not be included in the Docker image.

ivo-ritense (approved):

teodor-ritense (approved):

PR #56, opened 2025-10-09
GH-339: Added index to JsonSchemaDocument's internalCaseStatusKey column

marijnritense:
Solves https://github.com/generiekzaakafhandelcomponent/gzac-issues/issues/339

### DIFF
--- a/backend/document/src/main/resources/config/liquibase/changelog/20251009-add-document-index-case-status.xml
+++ b/backend/document/src/main/resources/config/liquibase/changelog/20251009-add-document-index-case-status.xml
@@ -21,6 +21,7 @@
     <changeSet author="Ritense" id="1">
         <createIndex indexName="idx_json_schema_document_case_status_key" tableName="json_schema_document">
             <column name="internal_case_status_key"/>
+            <column name="case_definition_name"/>
         </createIndex>
 
     </changeSet>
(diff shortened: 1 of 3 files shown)


theo-ritense:
Note: this only fixes one half of the issue. The FE should not filter on `[null]` by default in the first place.

ThomasMinkeRitense:
> Note: this only fixes one half of the issue. The FE should not filter on `[null]` by default in the first place.

So your proposal is to remove the status filter from the FE search when the case type doesn't have any statuses?

backend/document/src/main/resources/config/liquibase/changelog/20251009-add-document-index-case-status.xml
```diff
@@ -0,0 +1,27 @@
...
+  ~ See the License for the specific language governing permissions and
+  ~ limitations under the License.
+  -->
+
+<databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+                   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+                   xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.6.xsd">
+
+    <changeSet author="Ritense" id="1">
+        <createIndex indexName="idx_json_schema_document_case_status_key" tableName="json_schema_document">
```
ivo-ritense, line 22: since the status is joined on a combination of `internal_case_status_key` and `case_definition_name`, shouldn't we add `case_definition_name` to the index?

ivo-ritense (approved):

PR #58, opened 2025-10-10
Prevent column updates

ivo-ritense:
No sure if we want to do this. See https://github.com/generiekzaakafhandelcomponent/gzac-issues/issues/348


theo-ritense:
I don't think this is the proper solution, as the original issue only shows the tip of the iceberg. This probably happens way more often than the given example.

Having writable Transactions everywhere also limits the legroom of optimising performance by offloading to a read database for instance, and locking/isolation becomes a problem.

I was there when we designed the outbox module, but hindsight has made me wonder if we should revisit this part.

PR #61, opened 2025-10-10
Allow sending outbox events from read only transaction

ivo-ritense:
Potential solution to https://github.com/generiekzaakafhandelcomponent/gzac-issues/issues/348


backend/outbox/src/main/kotlin/com/ritense/outbox/ValtimoOutboxService.kt
```diff
@@ -39,8 +42,13 @@ open class ValtimoOutboxService(
     private val objectMapper: ObjectMapper,
     private val userProvider: UserProvider,
     private val cloudEventSource: String,
+    transactionManager: PlatformTransactionManager,
```
theo-ritense, line 45:
```suggestion
    private transactionManager: PlatformTransactionManager,
```

ivo-ritense: Since this doesn't have to be used as val, we don't need private

marijnritense (approved):

backend/outbox/src/main/kotlin/com/ritense/outbox/ValtimoOutboxService.kt
theo-ritense: You're right.

PR #66, opened 2025-10-14
GH-348: Improve readonly behavior

ivo-ritense:
Part of https://github.com/generiekzaakafhandelcomponent/gzac-issues/issues/348

### DIFF
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -177,26 +177,36 @@
     "dist/valtimo/access-control": {
       "name": "@valtimo/access-control",
       "version": "13.0.1",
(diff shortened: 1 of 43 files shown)


frontend/package-lock.json
theo-ritense, line 1: @ivo-ritense I don't think the Valtimo version for FE should be changed to 13.0.1 when merging to 12.19.0.

Should this file be changed at all?

ivo-ritense: I'll revert the file, but it shouldn't matter.

marijnritense (approved):

PR #76, opened 2025-10-16
GH-330: Unify outbox events

ivo-ritense:
https://github.com/generiekzaakafhandelcomponent/gzac-issues/issues/330


Klaas-Ritense (approved):

backend/case/src/main/kotlin/com/ritense/document/listener/DocumentEventListener.kt
```diff
@@ -1,48 +0,0 @@
...
-import com.ritense.document.domain.event.CaseCreatedEvent
-import com.ritense.document.domain.event.CaseUnassignedEvent
-import com.ritense.document.domain.event.DocumentCreatedEvent
-import com.ritense.document.event.DocumentAssigneeChangedEvent
-import com.ritense.document.event.DocumentUnassignedEvent
-import com.ritense.valtimo.web.sse.service.SseSubscriptionService
-import org.springframework.context.event.EventListener
-import org.springframework.transaction.event.TransactionalEventListener
-
-class DocumentEventListener(
```
Klaas-Ritense, line 29: > While testing I noticed that the task-update events where still triggered twice because of the `TaskUpdateListener`. I think that class can now be removed

After closer inspection, the task-outbox-events where never connected to SSE

PR #87, opened 2025-10-22
IKO: [BE] widget conditions & widget divider

Klaas-Ritense:
https://github.com/generiekzaakafhandelcomponent/atlas-internal/issues/104
https://github.com/Integraal-Klant-en-Objectbeeld/iko-issues/issues/128


backend/widget/src/main/resources/config/liquibase/changelog/20250704-add-widget-table.xml
```diff
@@ -90,18 +90,15 @@
         </createTable>
     </changeSet>
 
-    <changeSet id="3" author="Ritense">
```
ivo-ritense, line 93: I'm not sure why the old changeset has to be removed? Did you mean to copy them?

backend/widget/src/main/kotlin/com/ritense/widget/service/WidgetService.kt
```diff
@@ -51,6 +56,21 @@ class WidgetService(
     fun findById(id: UUID): Widget? =
         widgetRepository.findById(id).getOrNull()
 
+    fun filterWidgetsOnDisplayConditions(widgets: List<Widget>, properties: Map<String, Any>): List<Widget> {
+        return widgets.filter { widget ->
+            widget.displayConditions.all {
+                it.isValid { valuePath: String ->
+                    valueResolverService.resolveValues(
```
ivo-ritense: Looks like it would work, but every condition does another call to the valueResolverService which could mean lots of db queries when it's not needed. It there a good way of grouping them?

backend/widget/src/main/resources/config/liquibase/changelog/20250704-add-widget-table.xml
Klaas-Ritense: The changesets that I removed/changed here also exist in the case module. I copied them here without changing the tableName

backend/widget/src/main/kotlin/com/ritense/widget/service/WidgetService.kt
Klaas-Ritense: I've grouped them now. Also for the case widgets

ivo-ritense:
I tested this together with the FE from https://github.com/valtimo-platform/valtimo/pull/96 and will leave all purely FE comments there.

ivo-ritense:
When adding conditions to widgets, they no longer show up in the list for the admin screen either.

EDIT: nevermind. this is a frontend issue. The non-management url is called form the management screen.

ivo-ritense (approved):

PR #117, opened 2025-10-31
Fixed widget issues and style

ivo-ritense:
Fixes these stories:
- https://github.com/generiekzaakafhandelcomponent/gzac-issues/issues/388
- https://github.com/generiekzaakafhandelcomponent/gzac-issues/issues/367

### DIFF
--- a/frontend/projects/valtimo/layout/src/lib/components/widget-formio/widget-formio.component.scss
+++ b/frontend/projects/valtimo/layout/src/lib/components/widget-formio/widget-formio.component.scss
@@ -24,3 +24,20 @@
 .error-message {
   text-align: center;
 }
+
+.formio-widget {
+  padding: 24px;
+
+  &__header {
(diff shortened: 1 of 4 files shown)


frontend/projects/valtimo/layout/src/lib/components/widget-formio/widget-formio.component.scss
```diff
@@ -24,3 +24,20 @@
 .error-message {
   text-align: center;
 }
+
+.formio-widget {
```
ivo-ritense, line 28: I fixed this style because this way it's in line with the other widgets (and I think it's better). We might consider it a breaking change though if people took the lack of padding into account in their forms.

Here's an example with a table, (minimal) formIO form and custom component below each other.

Before:

Image

Now: 

Image

mbritense (approved):

PR #118, opened 2025-11-03
Bugfix: iko use query parameters

Klaas-Ritense:
https://github.com/Integraal-Klant-en-Objectbeeld/iko-issues/issues/154


backend/iko/src/main/kotlin/com/ritense/iko/helper/MapHelper.kt
```diff
@@ -0,0 +1,34 @@
...
+
+object MapHelper {
+
+    fun <T> Map<T, Any?>.deepMerge(other: Map<T, Any?>): Map<T, Any?> =
+        this.toMutableMap().apply {
+            other.forEach { (key, value) ->
+                val existing = this[key]
+                if (existing is Map<*, *> && value is Map<*, *>) {
+                    @Suppress("UNCHECKED_CAST")
+                    this[key] = (existing as Map<String, Any>).deepMerge(value as Map<String, Any>)
```
ivo-ritense: I'm not sure if this can go wrong on any type other than. `Map<String, Any>`, why not `Map<Any?, Any?>`? What would happen if you try to merge a `Map<Int, String>` with a `Map<String, LIst<String>>` in the properties? Is there a way to lock this down?

Klaas-Ritense: Ive changed it

marijnritense (approved):

backend/iko/src/main/kotlin/com/ritense/iko/helper/MapHelper.kt
ivo-ritense: It's still not typesafe i think, but we'll figure that out later if we have more use cases.

ivo-ritense (approved):

PR #122, opened 2025-11-04
story: add building blocks tests

mbritense:
-

### DIFF
--- a/backend/building-block/src/test/kotlin/com/ritense/buildingblock/web/rest/BuildingBlockDocumentDefinitionResourceIT.kt
+++ b/backend/building-block/src/test/kotlin/com/ritense/buildingblock/web/rest/BuildingBlockDocumentDefinitionResourceIT.kt
@@ -8,7 +8,7 @@
  * https://joinup.ec.europa.eu/collection/eupl/eupl-text-eupl-12
  *
  * Unless required by applicable law or agreed to in writing, software
- * distributed under the License is distributed on an "AS IS" BASIS,
+ * distributed under the License is distributed on an "AS IS" basis,
  * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
  * See the License for the specific language governing permissions and
  * limitations under the License.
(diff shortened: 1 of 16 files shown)


backend/building-block/src/test/kotlin/com/ritense/buildingblock/web/rest/BuildingBlockManagementResourceIT.kt
```diff
@@ -0,0 +1,341 @@
...
+        mockMvc.put("$base/{k}/version/{v}", key, version) {
+            contentType = MediaType.APPLICATION_JSON
+            content = objectMapper.writeValueAsBytes(mapOf("title" to "irrelevant"))
+        }.andExpect { status { isNotFound() } }
+    }
+
+    @Test
+    @WithMockUser
+    @DisplayName("should create then get then update then list definitions")
+    fun shouldCreateThenGetThenUpdateThenListDefinitions() {
```
ivo-ritense: Not sure what this test adds apart from doing the previous tests together

backend/building-block/src/test/resources/config/application-mysql.yml
```diff
@@ -0,0 +1,16 @@
+spring:
+    datasource:
+        driver-class-name: com.mysql.cj.jdbc.Driver
+        url: jdbc:mysql://localhost:3363/building-block-test
```
ivo-ritense, line 4: This is fine, but also add a comment to the docker-compose-base list so we don't use it twice in the future.

backend/building-block/src/test/kotlin/com/ritense/buildingblock/web/rest/BuildingBlockDocumentDefinitionResourceIT.kt
```diff
@@ -0,0 +1,132 @@
...
+    @MockitoBean
+    lateinit var repository: BuildingBlockJsonSchemaDocumentDefinitionRepository
+
+    private val base = "/api/management/v1/building-block"
+    private val key = "bb-key"
+    private val version = "1.0.0"
+
+    @Test
+    @WithMockUser
+    @DisplayName("should return 404 when document definition is missing")
```
ivo-ritense: This is fine but usually we use the kotlin ``fun `should return 404 when document definition is missing`()`` instead

ivo-ritense (approved):

PR #154, opened 2025-11-20
story: building block version switch

mbritense:
-

PR #155, opened 2025-11-20
Link building block to case

ivo-ritense:


backend/building-block/src/main/kotlin/com/ritense/buildingblock/web/rest/BuildingBlockManagementResource.kt
```diff
@@ -65,6 +67,14 @@ class BuildingBlockManagementResource(
         return ResponseEntity.ok(savedDto)
     }
 
+    @GetMapping("/{key}/version")
+    fun getBuildingBlockVersions(
```
ivo-ritense: Has conflict with https://github.com/valtimo-platform/valtimo/pull/154/files#diff-689bd88cfbd7bdaad4168af20481a7c5732a944cb4914e9516bf5e32192f40dc. We should probably use that version and adjust in this branch.

frontend/projects/valtimo/building-block-management/src/lib/building-block-management-routing.ts
```diff
@@ -43,7 +46,10 @@ const routes: Routes = [
   },
   {
     path: `building-block-management/building-block/:buildingBlockDefinitionKey/version/:buildingBlockDefinitionVersionTag/${BUILDING_BLOCK_MANAGEMENT_TABS.PROCESSES}/:processDefinitionKey`,
-    component: ProcessManagementBuilderComponent,
+    loadComponent: () =>
```
mbritense, line 49: Waarom is dit lazy loaded gemaakt?

frontend/projects/valtimo/process-link/src/lib/components/select-building-block/select-building-block.component.scss
```diff
@@ -0,0 +1,24 @@
+.loading-container {
+  display: flex;
+  justify-content: center;
+  padding: 2rem 0;
```
mbritense: We us px in css, so 1rem = 16px

frontend/projects/valtimo/building-block-management/src/lib/building-block-management-routing.ts
ivo-ritense: Because there's a circular dependency we break this way. Or we would get this error:

Uncaught ReferenceError: can't access lexical declaration 'ProcessLinkStateService' before initialization

mbritense (approved):

PR #161, opened 2025-11-21
fix: set correct case version tags on call activities on import

mbritense:
closes https://github.com/orgs/generiekzaakafhandelcomponent/projects/5/views/3?sliceBy%5Bvalue%5D=SPRINT+W47-W48&pane=issue&itemId=140672209&issue=generiekzaakafhandelcomponent%7Cgzac-issues%7C432

### DIFF
--- a/backend/core/src/main/java/com/ritense/valtimo/service/OperatonProcessService.java
+++ b/backend/core/src/main/java/com/ritense/valtimo/service/OperatonProcessService.java
@@ -32,6 +32,14 @@ import com.ritense.authorization.Action;
 import com.ritense.authorization.AuthorizationContext;
 import com.ritense.authorization.AuthorizationService;
 import com.ritense.authorization.request.EntityAuthorizationRequest;
+import com.ritense.valtimo.contract.case_.CaseDefinitionId;
+import com.ritense.valtimo.contract.config.ValtimoProperties;
+import com.ritense.valtimo.event.ProcessDefinitionDeleted;
+import com.ritense.valtimo.exception.FileExtensionNotSupportedException;
+import com.ritense.valtimo.exception.NoFileExtensionFoundException;
+import com.ritense.valtimo.exception.ProcessDefinitionNotFoundException;
+import com.ritense.valtimo.exception.ProcessNotDeployableException;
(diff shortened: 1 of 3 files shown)


backend/core/src/main/java/com/ritense/valtimo/service/OperatonProcessService.java
```diff
@@ -685,32 +689,34 @@ private void setProcessesVersionTag(BpmnModelInstance bpmnModel, CaseDefinitionI
...
-                ) {
-                    callActivity.setOperatonCalledElementBinding("versionTag");
-                    callActivity.setOperatonCalledElementVersionTag(OPERATON_CASE_DEFINITION_VERSION_TAG_PREFIX + caseDefinitionId);
-                }
+            CaseDefinitionId existingCaseDefinitionId =
+                CaseDefinitionId.fromProcessVersionTag(existingVersionTag);
+
+            // return if binding is null, or if there is already a version tag and it's not a case definition version tag
+            if (binding != null && (existingCaseDefinitionId == null && existingVersionTag != null)) {
+                return;
```
Klaas-Ritense, line 711: The 'return' will stop the for-loop. So this can go wrong when there are multiple call-activities in the BPMN

mbritense: I looked it up, but the return only has effect inside the lambda function for this particular forEach, but the one methods run after for the forEach will continue to work. Can you recheck this?

Klaas-Ritense: you are right

Klaas-Ritense:
Test findings
ERROR scenario:
 1. Go to /Admin/Cases/'Bezwaar'
 2. Go to Processes
 3. Create random process
 4. Export the bpmn process
 5. Import this process in the 'global' /Admin/Processes
 6. Notice the process isn't added to the list
 7. Go back to /Admin/Cases/'Bezwaar'/Processes
 8. Notice a 500 error. Its no longer possible to edit processes
image

Klaas-Ritense (approved):

PR #169, opened 2025-11-26
story: add compact widgets

mbritense:
Testing steps

- [ ] Go to case admin/IKO admin
- [ ] Create a widget (steps after width should not be visible)
- [ ] Select Fields/Collection/Table widget
- [ ] Steps should now include a Density step
- [ ] If any other widget is selected, Density step should disappear
- [ ] Finish configuring the widget
- [ ] Check compact widgets match the [design](https://www.figma.com/design/CV89FyTwuQx7klWgLJ92ws/Valtimo?node-id=11672-38214)
- [ ] Check that header widget is still displayed correctly
- [ ] Configuration for the header widget should not include Density step as it is always compact

Test BSN for IKO: 999990755


marijnritense (approved):

mbritense:
Test OK, the only thing I'm missing is that I expect the density to be listed in these two list views 
image
image

PR #196, opened 2025-12-08
story: building block process management

mbritense:
- create new building block process definition from ui
- mark different building block process definition as main from ui
- delete building block process definition from ui

### DIFF
--- a/backend/building-block/src/main/kotlin/com/ritense/buildingblock/service/BuildingBlockDefinitionProcessDefinitionService.kt
+++ b/backend/building-block/src/main/kotlin/com/ritense/buildingblock/service/BuildingBlockDefinitionProcessDefinitionService.kt
@@ -262,6 +262,47 @@ class BuildingBlockDefinitionProcessDefinitionService(
         return mainLink.processDefinitionKey
     }
 
+    @Transactional
+    fun setMainProcessDefinition(
+        buildingBlockDefinitionKey: String,
+        buildingBlockDefinitionVersionTag: String,
+        processDefinitionId: String
+    ) {
+        denyAuthorization()
+
+        val buildingBlockDefinitionId = BuildingBlockDefinitionId.of(
+            buildingBlockDefinitionKey,
+            buildingBlockDefinitionVersionTag
+        )
+
+        buildingBlockDefinitionChecker.assertCanUpdateBuildingBlockDefinition(buildingBlockDefinitionId)
+
+        val targetProcessDefinitionId = ProcessDefinitionId.of(processDefinitionId)
+
+        val targetLink = processDefinitionBuildingBlockDefinitionRepository
+            .findByIdBuildingBlockDefinitionIdAndIdProcessDefinitionId(
+                buildingBlockDefinitionId,
+                targetProcessDefinitionId
+            ) ?: return
+
+        if (targetLink.main) {
+            ensureOnlyOneMainLink(buildingBlockDefinitionId, targetProcessDefinitionId)
+            return
+        }
+
+        val updatedMainLink = ProcessDefinitionBuildingBlockDefinition(
+            ProcessDefinitionBuildingBlockDefinitionId(
+                targetProcessDefinitionId,
+                buildingBlockDefinitionId
+            ),
+            true
+        )
+
+        processDefinitionBuildingBlockDefinitionRepository.save(updatedMainLink)
+
+        ensureOnlyOneMainLink(buildingBlockDefinitionId, targetProcessDefinitionId)
+    }
+
     private fun findExistingLink(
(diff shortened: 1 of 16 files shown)


backend/building-block/src/main/kotlin/com/ritense/buildingblock/service/BuildingBlockDefinitionProcessDefinitionService.kt
```diff
@@ -262,10 +262,95 @@ class BuildingBlockDefinitionProcessDefinitionService(
...
+        val targetProcessDefinitionId = ProcessDefinitionId.of(processDefinitionId)
+
+        val targetLink = processDefinitionBuildingBlockDefinitionRepository
+            .findByIdBuildingBlockDefinitionIdAndIdProcessDefinitionId(
+                buildingBlockDefinitionId,
+                targetProcessDefinitionId
+            ) ?: return
+
+        if (targetLink.main) {
+            ensureOnlyOneMainLink(buildingBlockDefinitionId, targetProcessDefinitionId)
```
marijnritense: Why are we validating this here? The one on line 303 should be enough.

backend/building-block/src/main/kotlin/com/ritense/buildingblock/service/BuildingBlockDefinitionProcessDefinitionService.kt
```diff
@@ -262,10 +262,95 @@ class BuildingBlockDefinitionProcessDefinitionService(
...
+
+        val targetProcessDefinitionId = ProcessDefinitionId.of(processDefinitionId)
+
+        val targetLink = processDefinitionBuildingBlockDefinitionRepository
+            .findByIdBuildingBlockDefinitionIdAndIdProcessDefinitionId(
+                buildingBlockDefinitionId,
+                targetProcessDefinitionId
+            ) ?: return
+
+        if (targetLink.main) {
```
ivo-ritense, line 288: What is this meant to do? When is is already the main process make sure it's the only one? But shouldn't it already be the only one anyway?

backend/building-block/src/main/kotlin/com/ritense/buildingblock/service/BuildingBlockDefinitionProcessDefinitionService.kt
```diff
@@ -262,10 +262,95 @@ class BuildingBlockDefinitionProcessDefinitionService(
         return mainLink.processDefinitionKey
     }
 
+    @Transactional
+    fun setMainProcessDefinition(
```
marijnritense, line 266: I might be misunderstanding something. What I expect to see is the retrieval of the process that is currently set to main, that one getting updated to no longer be main, and the target process definition being set to main. Both should already have records in the table.

backend/building-block/src/main/kotlin/com/ritense/buildingblock/service/BuildingBlockDefinitionProcessDefinitionService.kt
```diff
@@ -262,10 +262,95 @@ class BuildingBlockDefinitionProcessDefinitionService(
...
+            true
+        )
+
+        processDefinitionBuildingBlockDefinitionRepository.save(updatedMainLink)
+
+        ensureOnlyOneMainLink(buildingBlockDefinitionId, targetProcessDefinitionId)
+    }
+
+    @Transactional
+    fun deleteProcessDefinitionForBuildingBlock(
```
marijnritense, line 314: `ProcessDefinitionDeletedEventListener` can handle deleting process links. There's even TODOs there specifically to handle building blocks. You could expand `OperatonProcessService.deleteProcessDefinition` to publish the `ProcessDefinitionDeletedEvent`.

ivo-ritense (approved):

backend/building-block/src/main/kotlin/com/ritense/buildingblock/service/BuildingBlockDefinitionProcessDefinitionService.kt
```diff
@@ -262,10 +262,97 @@ class BuildingBlockDefinitionProcessDefinitionService(
...
+
+        if (targetLink.main) {
+            return
+        }
+
+        val currentMainLink = processDefinitionBuildingBlockDefinitionRepository
+            .findByIdBuildingBlockDefinitionIdAndMain(buildingBlockDefinitionId, true)
+
+        processDefinitionBuildingBlockDefinitionRepository.save(
+            ProcessDefinitionBuildingBlockDefinition(
```
ivo-ritense, line 296: For now this might be fine. If we extend this entity a copy (like) function might be nice because it won't break functions like this.

PR #225, opened 2025-12-16
story: building block zaak instance link and zaak type link dependencies warning

mbritense:
Closes https://github.com/orgs/generiekzaakafhandelcomponent/projects/5?pane=issue&itemId=3730201348&issue=generiekzaakafhandelcomponent%7Catlas-internal%7C161

### DIFF
--- a/backend/plugin/src/main/kotlin/com/ritense/plugin/annotation/Plugin.kt
+++ b/backend/plugin/src/main/kotlin/com/ritense/plugin/annotation/Plugin.kt
@@ -15,7 +15,6 @@
  */
 
 package com.ritense.plugin.annotation
-import com.ritense.plugin.domain.PluginDependency
 
 @Target(AnnotationTarget.CLASS)
 @Retention(AnnotationRetention.RUNTIME)
@@ -23,5 +22,5 @@ annotation class Plugin(
     val key: String,
     val title: String,
     val description: String,
-    val dependencies: Array<PluginDependency> = []
+    val dependencies: Array<String> = []
 )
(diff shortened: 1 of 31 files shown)


backend/plugin/src/main/kotlin/com/ritense/plugin/service/PluginService.kt
```diff
@@ -479,6 +480,26 @@ class PluginService(
...
+
+        val mergedDependencies = definitions
+            .mapNotNull { definition ->
+                val pluginClass = Class.forName(definition.fullyQualifiedClassName)
+                pluginClass.getAnnotation(Plugin::class.java)
+            }
+            .flatMap { it.dependencies.asList() }
+            .toSet()
+
+        return PluginDefinitionsWithDependenciesDto(
```
marijnritense, line 507: I think this works well. Do you think it's likely for people to want to know why a certain dependency is in here? If you do, maybe it's worth it to instead have a list of Plugin keys with their dependencies, rather than one list with all plugin keys and one list with all dependencies. If we don't consider this right now, it might end up in a breaking change later on (since the REST api is tied to this DTO).

backend/plugin/src/main/kotlin/com/ritense/plugin/annotation/Plugin.kt
```diff
@@ -15,11 +15,13 @@
...
 package com.ritense.plugin.annotation
+import com.ritense.plugin.domain.PluginDependency
 
 @Target(AnnotationTarget.CLASS)
 @Retention(AnnotationRetention.RUNTIME)
 annotation class Plugin(
     val key: String,
     val title: String,
     val description: String,
+    val dependencies: Array<PluginDependency> = []
```
marijnritense: Since other people might want to make their own dependencies (we allow users to create their own plugins and publish them), I don't think this should be an enum.

marijnritense (approved):

PR #236, opened 2025-12-18
Do not delete previous process definition when editing draft

Klaas-Ritense:
https://github.com/generiekzaakafhandelcomponent/atlas-internal/issues/164


backend/core/src/main/java/com/ritense/valtimo/service/OperatonProcessService.java
```diff
@@ -556,14 +551,10 @@ public DeploymentWithDefinitions deploy(
...
-
-                if (caseDefinitionId != null) {
-                    decisionDefinitionQuery.versionTag(OPERATON_CASE_DEFINITION_VERSION_TAG_PREFIX + caseDefinitionId);
-                }
-
-                DecisionDefinition decisionDefinition = decisionDefinitionQuery.singleResult();
+                DecisionDefinition decisionDefinition = repositoryService.createDecisionDefinitionQuery()
+                    .decisionDefinitionKey(decisionDefinitionKey)
+                    .versionTag(OPERATON_CASE_DEFINITION_VERSION_TAG_PREFIX + caseDefinitionId)
+                    .singleResult();
```
Klaas-Ritense, line 563: No logic was changed here. See line 549

backend/core/src/main/java/com/ritense/valtimo/service/OperatonProcessService.java
```diff
@@ -556,14 +556,10 @@ public DeploymentWithDefinitions deploy(
...
 
                 if (decisionDefinition != null) {
                     repositoryService.deleteDeployment(decisionDefinition.getDeploymentId());
```
Klaas-Ritense, line 566: The old DMN is still deleted on every update. This will not result in problems because DMNs have no running instances

ivo-ritense (approved):

ivo-ritense:
The issue with this is that the process might not run successfully because some plugin actions may assume they are linked to a case definition. Regardlesss, this is an improvement because the chances of unexpected data loss go down drastically.

PR #259, opened 2026-01-06
IKO ADP per tab

Klaas-Ritense:
https://github.com/Integraal-Klant-en-Objectbeeld/iko-issues/issues/215


frontend/projects/valtimo/iko/src/lib/components/iko-management-details/components/tab-detail-modal/iko-management-tab-details-modal.component.ts
```diff
@@ -114,6 +123,10 @@ export class IkoManagementTabDetailsModalComponent {
   public get type(): AbstractControl<string> {
     return this.form.get('type') as AbstractControl<string>;
   }
+  public get properties(): FormGroup | null {
+    const properties = this.form.get('properties');
+    return !properties ? null : (properties as FormGroup);
+  }
```
sofiaIvarsRitense, line 128: I think this would look better like this:
```
public get properties(): FormGroup | null {
  return this.form.get('properties') as FormGroup | null;
}
```

Klaas-Ritense: changed

mbritense:
Tested, code looks good and works well. What I'm missing is having an overview of the extra configured properties for a tab in the list view (so it's visible before opening the edit modal)

mbritense (approved):

PR #268, opened 2026-01-09
feat: case definition can only be finalized if all building blocks it uses are final

mbritense:
-


backend/case/src/main/kotlin/com/ritense/case/service/CaseDefinitionService.kt
```diff
@@ -240,10 +243,17 @@ class CaseDefinitionService(
 
     fun finalizeCaseDefinition(caseDefinitionId: CaseDefinitionId): CaseDefinition {
         denyManagementOperation()
+
+        val check = isCaseDefinitionFinalizable(caseDefinitionId)
+        require(check.finalizable) {
+            "Failed to finalize case-definition. Case-definition with id: '$caseDefinitionId' cannot be made definitive. Reason: '${check.code}'."
```
ivo-ritense, line 250: I think in the future we might want to collect all code at the same time, so users will not have to find the next issue all the time. For now i think it doesn't matter much.

ivo-ritense (approved):

PR #270, opened 2026-01-09
Prevent log errors on empty DB

### LINKED ISSUE #468: ERROR in ch.qos.logback.classic.db.DBAppender[VALTIMODB] - problem appending event org.postgresql.util.PSQLException: ERROR: relation "logging_event" does not exist
## Affected version
*Specify in which version of the application the issue is found?*
13.7.0

## Expected behavior
*What should happen?*
The application starts up without errors

## Current behavior
*What happens instead of the expected behavior?*
Logs are being inserted into the logs table. But that doesn't exist right away. So a lot of errors.

## Possible solution (optional)
*What would be a possible fix?*

## Steps to reproduce
*Provide an unambiguous set of steps to reproduce this bug.*
*Include code to reproduce, if relevant*
1. Start up GZAC with a fresh database
2.Notice logs

Image

ivo-ritense:
Fixes: https://github.com/generiekzaakafhandelcomponent/gzac-issues/issues/468

Added filter that rejects log lines for being stored in db until the table exists.


backend/logging/src/main/kotlin/com/ritense/logging/LogDbReadyFilter.kt
Klaas-Ritense, line 1: @ivo-ritense missing release notes

Klaas-Ritense (approved):

PR #276, opened 2026-01-12
Iframe widget

ThomasMinkeRitense:


backend/case/src/main/kotlin/com/ritense/case_/configuration/CaseWidgetAutoConfiguration.kt
Klaas-Ritense, line 1: 1. There are backend tests failing
2. The frontend code failed to compile

PR #299, opened 2026-01-20
feat: fix nested building blocks zgw plugin actions

mbritense:
Also adds zgw actions to test building blocks, removes the continuous sync option since the backend does not support this yet


backend/zgw/zaken-api/src/main/kotlin/com/ritense/zakenapi/ZakenApiPlugin.kt
```diff
@@ -302,10 +303,14 @@ class ZakenApiPlugin(
             "com.ritense.document.domain.impl.JsonSchemaDocument" to documentId.toString(),
         ) {
             logger.debug { "Starting creation of zaak of zaaktype with URL '$zaaktypeUrl' for document with id '$documentId'" }
-            val zaakInstanceLink = zaakInstanceLinkRepository.findByDocumentId(documentId)
+            val existingZaakUrl = try {
+                zaakUrlProvider.getZaakUrl(documentId)
+            } catch (e: ZaakInstanceLinkNotFoundException) {
+                null
```
marijnritense, line 309: We should never catch an exception and then do nothing. Can you add a log line here?

marijnritense: Never mind, I see this is coved a bit later.

marijnritense (approved):

PR #318, opened 2026-01-26
Updated `SchedulerAutoConfiguration` to configure `JdbcTemplateLockProvider` with specified timezone defaulting to UTC

maarten-ritense:
### Describe the changes
Link to the related Github issue: https://github.com/generiekzaakafhandelcomponent/gzac-issues/issues/540

Updated the LockProvider bean configuration to use the configuration builder which allows specifying the timezone. 

Specify the code branch location: `story/gh-540_shedlock-lockprovider-config`

**Relevant comments:**
> 

### Breaking changes

### Documentation

- [ ] Yes
- [x] Not applicable

### Tests
- [ ] Yes
- [x] Not applicable

- [ ] Yes
- [x] Not applicable

### Security
- [ ] Yes
- [x] Not applicable

- [ ] Yes
- [x] Not applicable

- [ ] Yes
- [x] Not applicable

### Dependencies
- [ ] Yes
- [x] Not applicable

## Summary by CodeRabbit

* **Bug Fixes**
  * Scheduler lock time zone is validated and now defaults to UTC when unspecified, preventing invalid time zone configuration.

* **Chores**
  * Lock provider creation is now conditional to avoid duplicate instances and supports deployment-configurable time zone.

* **Documentation**
  * Release notes updated to reflect the new configurable/default time zone behavior.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/backend/core/src/main/java/com/ritense/valtimo/autoconfigure/SchedulerAutoConfiguration.java
+++ b/backend/core/src/main/java/com/ritense/valtimo/autoconfigure/SchedulerAutoConfiguration.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2015-2024 Ritense BV, the Netherlands.
+ * Copyright 2015-2026 Ritense BV, the Netherlands.
  *
  * Licensed under EUPL, Version 1.2 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


backend/core/src/main/java/com/ritense/valtimo/autoconfigure/SchedulerAutoConfiguration.java
```diff
@@ -33,6 +35,11 @@ public class SchedulerAutoConfiguration {
     @Order(HIGHEST_PRECEDENCE + 13)
     @Bean
     public LockProvider lockProvider(DataSource dataSource) {
-        return new JdbcTemplateLockProvider(dataSource);
+        return new JdbcTemplateLockProvider(
+            JdbcTemplateLockProvider.Configuration.builder()
+                .withJdbcTemplate(new JdbcTemplate(dataSource))
+                .withTimeZone(TimeZone.getDefault())
+                .build()
+        );
```
Klaas-Ritense, line 57: This code runs **before** `ValtimoApplicationReadyEventListener` is executed.  
As a result, `TimeZone.getDefault()` here, is resolved **before** the JVM default is switched to UTC, and therefore ends up using the local system timezone (e.g. `Europe/Amsterdam` in my case), not UTC.

This can lead to incorrect ShedLock timestamp interpretation once the default timezone is changed.

Would it be possible to:
- set the lock provider explicitly to UTC using  
  `.withTimeZone(TimeZone.getTimeZone("UTC"))`, **or**
- make the timezone configurable in the same way as in `ValtimoApplicationReadyEventListener`?

Both options would avoid relying on the mutable JVM default timezone during startup and make the behavior deterministic.

maarten-ritense: Sure, I can have a look what happens when UTC is used instead. Provided configuration was what made it work on my local system and in the stack the TU uses.

maarten-ritense: Adjusted the bean so it now also accepts `timezone` value from configuration which defaults to UTC if not provided.

Klaas-Ritense (approved):

Klaas-Ritense:
@maarten-ritense An integration test is failing: `ProcessInstanceEventHandlerIntTest`

PR #348, opened 2026-01-30
cicd: Make separate test/prod backend images, use test images in PR test envs 

sjorsholtrop-ritense:
Updated the backend CI pipeline to build both production and test Docker images in parallel.

Build Matrix: Added production vs test matrix to backend build jobs.
test builds run with -PincludeDev=true to include dev/test configurations.
production builds run with -PincludeDev=false.
Test Environments: backend_ci now explicitly deploys the test image variant to PR environments.
Tagging: Updated Docker tags to distinguish between build types (e.g., commit-test-abc vs commit-production-abc).


RemcoRitense (approved):
lgtm

PR #352, opened 2026-01-30
fix: use building blocks in independent processes

### LINKED ISSUE #547: [FE+BE?] Building block process links support in independent processes
Within the context of an independent proces (not under case or building block), doc: is not supported and it must be verified whether filling in pv here works.

Also, it says Dossier/Case on the left hand side of the input mapping and the right hand side of the output mapping, but it should say 'Process'

mbritense:
closes https://github.com/generiekzaakafhandelcomponent/gzac-issues/issues/547


backend/building-block/src/main/kotlin/com/ritense/buildingblock/processlink/service/BuildingBlockCallActivityListener.kt
```diff
@@ -57,16 +57,28 @@ class BuildingBlockCallActivityListener(
             // Check if we're inside a parent building block (nested building block scenario)
             val parentBuildingBlockInstance = findParentBuildingBlockInstance(execution)
 
+            // Determine if this is an independent process (no case, no parent building block)
+            val isIndependentProcess = parentBuildingBlockInstance == null && execution.businessKey == null
```
ivo-ritense: Not sure if the business key is always null for independent processes

ivo-ritense (approved):

teodor-ritense (approved):

PR #357, opened 2026-02-03
GH-545: Call depth thresholds for processes

marijnritense:
When a process starts another process, which starts another process (etc) via call activity, you can accidentally introduce a loop. This feature detects when such a loop happens through the local `VTM_callDepth` variable. This is stored inside of an execution and is calculated when a new process start event is triggered. When that call depth exceeds a threshold (configurable), a warning is emitted.

https://github.com/generiekzaakafhandelcomponent/gzac-issues/issues/545


ThomasMinkeRitense:
- Isn't 100 a little high as a reasonable treshold? I'd say 10 is already higher than I'd expect
- If this loop is executed without wait states, the number of process instances increases pretty fast. Is it an option to add a 'sleep' when the treshold has been reached to lower the impact on the database?

ivo-ritense (approved):

PR #367, opened 2026-02-05
feat: form management for building blocks

mbritense:
closes https://github.com/orgs/generiekzaakafhandelcomponent/projects/5?pane=issue&itemId=153855417&issue=generiekzaakafhandelcomponent%7Catlas-internal%7C185


backend/form/build.gradle
```diff
@@ -93,4 +96,15 @@ dependencies {
     }
 }
 
+// Handle duplicate liquibase files from circular test dependency (building-block -> form)
```
ivo-ritense, line 99: This shouldn't be needed anymore after the liquibase changes i think.

ivo-ritense:
When the building block is final, it is still possible in the UI to upload/delete even though the backend blocks the actual changes. This is not very user friendly. The one thing that does change is the save button on the form detail screen that gets removed. this in inconsistent with other building block elements, where the button is still on the screen, but is disabled instead.

sofiaIvarsRitense (approved):

ivo-ritense:
When creating a new form in a building block, and you fill in the name in the modal:
- When you press enter -> modal closes and form is not created (wrong)
- When you press submit -> you are taken to the new form page (correct)

ivo-ritense:
After linking a form to a user task in a building block in the modeller the form name isn't displayed like it is for cases

Case:
image

Building block:
image

ivo-ritense (approved):

sofiaIvarsRitense (approved):

PR #392, opened 2026-02-17
Fix Task due date language

Klaas-Ritense:
https://github.com/generiekzaakafhandelcomponent/gzac-issues/issues/181


mbritense:
The language switching works, but when I set a due date here and save it, it saves in the task, but the last task list is not loaded with latest info, so when I click the task again I don't see the due date even though it is saved in the back-end: https://pr392.product-development.test.k8s.ritense.com/tasks?selectedTaskType=ImFsbCI%3D&params=eyJwYWdlIjowLCJzaXplIjoxMH0%3D&otherFilters=W10%3D

mbritense (approved):

PR #433, opened 2026-03-10
Removed unused cron job for task notifications

marijnritense:
Solves https://github.com/generiekzaakafhandelcomponent/gzac-issues/issues/322


sofiaIvarsRitense (approved):

backend/core/src/main/java/com/ritense/valtimo/operaton/task/service/impl/ReminderServiceImpl.java
```diff
@@ -35,6 +35,7 @@
 import net.javacrumbs.shedlock.spring.annotation.SchedulerLock;
 import org.springframework.scheduling.annotation.Scheduled;
 
+// TODO: Remove this altogether
```
mbritense: Why not remove it altogether now? I don't think that would go against the new breaking change policy

marijnritense: Done

mbritense (approved):

PR #438, opened 2026-03-10
Story 248 Team Management

Klaas-Ritense:
https://github.com/generiekzaakafhandelcomponent/atlas-internal/issues/248


1.png
marijnritense, line 1: Why are these images at root level?

documentation/features/teams/for-developers/README.md
```diff
@@ -0,0 +1,49 @@
+# Teams (for developers)
```
marijnritense, line 1: Weren't we going to skip writing these?

marijnritense:
Can you remove the role field for the team list?

backend/keycloak-iam/src/main/java/com/valtimo/keycloak/autoconfigure/KeycloakAutoConfiguration.java
```diff
@@ -84,9 +84,16 @@ public KeycloakUserManagementService keycloakUserManagementService(
         final KeycloakService keycloakService,
         @Value("#{'${spring.security.oauth2.client.registration.keycloakjwt.client-id:${valtimo.keycloak.client:}}'}") final String keycloakClientName,
         final UserCache userCache,
-        @Lazy final AuthorizationService authorizationService
+        @Lazy final AuthorizationService authorizationService,
+        @Lazy final TeamProvider teamProvider
```
marijnritense: Do all implementations always use the valtimo dependencies? If they do, they'll have the team module in their implementation. But I don't think this is always the case, some older projects probably don't. Can you make the teamprovider optional like in the case module?

backend/team/src/main/kotlin/com/ritense/team/service/TeamManagementServiceImpl.kt
```diff
@@ -0,0 +1,132 @@
...
+    fun findAll(
+        titleContains: String? = null,
+        pageable: Pageable = Pageable.unpaged()
+    ): Page<Team> {
+        val specification = if (titleContains != null) {
+            TeamRepositoryConfigSpecificationHelper.byTitleContains(titleContains)
+        } else {
+            Specification.unrestricted()
+        }
+        val teams = teamRepository.findAll(specification, pageable)
```
marijnritense, line 67: Shouldn't this use access control?

backend/team/src/main/kotlin/com/ritense/team/service/TeamService.kt
```diff
@@ -0,0 +1,132 @@
...
+            Specification.unrestricted()
+        }
+        val teams = teamRepository.findAll(specification, pageable)
+        teams.forEach { team -> team.users.size } // Initialize lazy collection
+        return teams
+    }
+
+    @Transactional(readOnly = true)
+    fun findById(key: String): Team? {
+        val team = teamRepository.findById(key).orElse(null)
```
marijnritense: Shouldn't this use access control?

backend/team/src/main/kotlin/com/ritense/team/service/TeamManagementServiceImpl.kt
```diff
@@ -0,0 +1,132 @@
...
+        pageable: Pageable = Pageable.unpaged()
+    ): Page<TeamUser> {
+        var specification = Specification.unrestricted<TeamUser>()
+        if (teamKey != null) {
+            specification = specification.and(TeamUserRepositoryConfigSpecificationHelper.byTeamKey(teamKey))
+        }
+        if (username != null) {
+            specification = specification.and(TeamUserRepositoryConfigSpecificationHelper.byUsername(username))
+        }
+        return teamUserRepository.findAll(specification, pageable)
```
marijnritense: Shouldn't this use access control?

backend/case/src/main/java/com/ritense/document/service/impl/JsonSchemaDocumentSearchService.java
```diff
@@ -446,11 +451,22 @@ private Predicate getAssigneeFilterPredicate(
...
+                var username = userManagementService.getCurrentUser().getUsername();
+                yield cb.equal(caseAssigneeIdColumn, username);
+            }
+            case TEAM -> {
+                var username = userManagementService.getCurrentUser().getUsername();
+                if (teamProvider == null) {
+                    throw new IllegalStateException();
+                }
+                var teamKeys = teamProvider.findTeamKeysByUsername(username);
+                yield cb.in(caseAssignedTeamKeyColumn).value(teamKeys);
```
marijnritense: What if the user isn't a part of any team? Then this would produce incorrect sql right?

backend/case/src/main/java/com/ritense/document/service/impl/JsonSchemaDocumentSearchService.java
```diff
@@ -446,11 +451,22 @@ private Predicate getAssigneeFilterPredicate(
...
-            case MINE -> cb.equal(caseAssigneeIdColumn, userId);
-            case OPEN -> cb.isNull(caseAssigneeIdColumn);
+            case MINE -> {
+                var username = userManagementService.getCurrentUser().getUsername();
+                yield cb.equal(caseAssigneeIdColumn, username);
+            }
+            case TEAM -> {
+                var username = userManagementService.getCurrentUser().getUsername();
+                if (teamProvider == null) {
+                    throw new IllegalStateException();
```
marijnritense: Can this exception be a bit more descriptive? If users see this in the logs they'll have no idea what went wrong.

backend/team/src/main/kotlin/com/ritense/team/service/TeamManagementServiceImpl.kt
```diff
@@ -0,0 +1,132 @@
...
+
+    fun create(team: Team): Team {
+        requirePermission(team, TeamActionProvider.CREATE)
+        return teamRepository.save(team)
+    }
+
+    fun update(key: String, title: String): Team {
+        val team = teamRepository.findById(key).orElseThrow { IllegalArgumentException("Team not found") }
+        requirePermission(team, TeamActionProvider.MODIFY)
+        team.title = title
```
marijnritense, line 94: How is the assigned team title updated for existing documents? Since the title is stored in there as well.

marijnritense:
~~FE wise, I think the only thing that needs to be fixed for the team management part is graying out options you're not allowed to do. So if you're not allowed to assign people to a team, or not allowed to delete any, I would expect the option to be grayed out.~~

Turns out this is because the results are being cached. This scenario will not happen in practice, it only occurs when you previously did have the permissions and they've now been removed. Even then, BE ensures the changes can't be made.

sofiaIvarsRitense (approved):

backend/case/src/main/java/com/ritense/document/domain/impl/JsonSchemaDocument.java
```diff
@@ -144,6 +144,12 @@ public class JsonSchemaDocument extends AbstractAggregateRoot<JsonSchemaDocument
     @Column(name = "assignee_full_name", columnDefinition = "varchar(255)")
     private String assigneeFullName;
 
+    @Column(name = "assigned_team_key", columnDefinition = "varchar(255)")
+    private String assignedTeamKey;
+
+    @Column(name = "assigned_team_title", columnDefinition = "varchar(255)")
+    private String assignedTeamTitle;
```
ivo-ritense, line 151: I don't understand why this is title and not name.

backend/case/src/main/java/com/ritense/document/web/rest/DocumentResource.java
```diff
@@ -50,6 +57,12 @@ public interface DocumentResource {
 
     ResponseEntity<Void> unassignHandlerFromDocument(UUID documentId);
 
+    @GetMapping("/v1/document/{document-id}/candidate-team")
+    ResponseEntity<Page<Team>> getCandidateTeams(
```
ivo-ritense: This is not in line with the structure of the file. Any reason to do it like this?

backend/core/src/main/kotlin/com/ritense/valtimo/operaton/dto/OperatonTaskDto.kt
```diff
@@ -41,13 +41,16 @@ data class OperatonTaskDto(
     val caseInstanceId: String?,
     val caseDefinitionId: String?,
     val suspended: Boolean,
-    val tenantId: String?
+    val tenantId: String?,
+    val assignedTeam: TeamDto? = null,
```
ivo-ritense, line 45: I'm not sure if this makes sense. Should this class reflect the operaton task? because the team is not part of that. It doubles up on the assignee information like this by having both `assignee` and `valtimoAssignee`. Do these fields get filled everywhere they are used? If not, does that matter? We do represent a false state of the task if we don't include these.

backend/core/src/main/resources/config/liquibase/initial-setup/task/changelog/20260317-create-table-task-team.xml
```diff
@@ -0,0 +1,44 @@
...
+
+    <changeSet id="1" author="Ritense">
+        <createTable tableName="task_team">
+            <column name="task_id" type="VARCHAR(64)">
+                <constraints primaryKey="true" nullable="false" primaryKeyName="pk_task_team"/>
+            </column>
+            <column name="team_key" type="VARCHAR(255)">
+                <constraints nullable="false"/>
+            </column>
+            <column name="team_title" type="VARCHAR(255)">
```
ivo-ritense, line 31: Why do we duplicate this data in this table instead of joining to the team table?

documentation/release-notes/13.x.x/13.22.0/front-end-migration.md
```diff
@@ -0,0 +1,53 @@
+# Front-end migration
```
ivo-ritense, line 1: Do we consider this a breaking change?

ivo-ritense: It's strange because the backend looks like it's set up to be modular. i.e. only support teams when the teams module has been included, but the frontend always needs the team module?

backend/case/src/main/java/com/ritense/document/domain/impl/JsonSchemaDocument.java
Klaas-Ritense: You mean assignedTeamName instead of assignedTeamTitle? The solution design calls it: title

backend/team/src/main/kotlin/com/ritense/team/domain/TeamUser.kt
```diff
@@ -0,0 +1,40 @@
...
+class TeamUser(
+    @EmbeddedId
+    val id: TeamUserId,
+
+    @MapsId("teamKey")
+    @ManyToOne(fetch = FetchType.LAZY)
+    @JoinColumn(name = "team_key")
+    val team: Team? = null
+) {
+    val username: String get() = id.username ?: ""
```
ivo-ritense: Not sure about this. it feels like a null like value in an interface where null is not allowed. Is it better in this case to return empty string and hope nothing goes wrong down the line, or should we throw an exception here?

backend/team/src/main/kotlin/com/ritense/team/service/TeamManagementServiceImpl.kt
```diff
@@ -0,0 +1,165 @@
...
+                Team::class.java,
+                TeamActionProvider.VIEW_LIST
+            )
+        )
+
+        if (titleContains != null) {
+            specification = specification.and(TeamRepositoryConfigSpecificationHelper.byTitleContains(titleContains))
+        }
+        val teams = teamRepository.findAll(specification, pageable)
+        teams.forEach { team -> team.users.size } // Initialize lazy collection
```
ivo-ritense: Not sure this is the best solution. if the teams have a large number of users, it will initialize these for no reason since we only want the number. Maybe we want another query or calculated value with a count()?

backend/team/src/main/kotlin/com/ritense/team/service/TeamManagementServiceImpl.kt
```diff
@@ -0,0 +1,165 @@
...
+    fun findAllTeamUsers(
+        teamKey: String? = null,
+        username: String? = null,
+        pageable: Pageable = Pageable.unpaged()
+    ): Page<TeamUser> {
+        var specification = Specification.unrestricted<TeamUser>()
+        if (teamKey != null) {
+            val team = teamRepository.findById(teamKey).orElse(null)
+
+            authorizationService.requirePermission(
```
ivo-ritense: You could use the private requirePermission here

backend/team/src/main/kotlin/com/ritense/team/web/rest/TeamResource.kt
```diff
@@ -0,0 +1,123 @@
...
+    fun getAllTeams(
+        @RequestParam(required = false) titleContains: String?,
+        pageable: Pageable
+    ): Page<TeamListResponseDto> {
+        return teamManagementService.findAll(titleContains, pageable).map { TeamListResponseDto.from(it) }
+    }
+
+    @GetMapping("/{key}")
+    fun getTeamById(@PathVariable key: String): TeamResponseDto {
+        val team = teamManagementService.findByKey(key) ?: throw IllegalArgumentException("Team not found")
```
ivo-ritense: This is not consistent with the NoSuchElementException used in TeamManagementServiceImpl

backend/team/src/main/kotlin/com/ritense/team/web/rest/TeamResource.kt
```diff
@@ -0,0 +1,123 @@
...
+    @PostMapping
+    @ResponseStatus(HttpStatus.CREATED)
+    fun createTeam(@Valid @RequestBody request: TeamCreateRequestDto): TeamResponseDto {
+        val team = teamManagementService.create(Team(key = request.key, title = request.title))
+        return TeamResponseDto.from(team)
+    }
+
+    @PutMapping("/{key}")
+    fun updateTeam(@PathVariable key: String, @Valid @RequestBody request: TeamUpdateRequestDto): TeamResponseDto {
+        require(request.key == key) { "Key in request does not match path variable" }
```
ivo-ritense: Strange in a way that we request this twice in the same request.

backend/team/src/main/kotlin/com/ritense/team/web/rest/TeamResource.kt
```diff
@@ -0,0 +1,123 @@
...
+        @PathVariable username: String
+    ) {
+        teamManagementService.removeUserFromTeam(username, teamKey)
+    }
+
+    @GetMapping("/{teamKey}/candidate-user")
+    fun getCandidateUsers(@PathVariable teamKey: String): List<ManageableUser> {
+        val teamMembers = teamManagementService.findAllTeamUsers(teamKey = teamKey)
+        val memberUsernames = teamMembers.content.map { it.username }.toSet()
+        return userManagementService.allUsers
```
ivo-ritense, line 121: How does this perform when we have a lot of users?

mbritense:
What's been changed and what to test:
---

## 1. Team Management (new module)

### CRUD operations
- **Teams list page** (`/teams`) — paginated table showing team name and member count
- **Create team** — modal with team name input
- **Edit team** — update team name
- **Delete team** — with confirmation modal
- **Team detail page** — shows team info and member list

### Team membership
- **View members** — paginated list showing full name and email
- **Add member** — modal with user selection (loads candidate users not already in the team)
- **Remove member** — remove a user from a team

### What to test:
- Create, edit, delete teams as admin
- Add/remove users to/from a team
- Verify the teams menu item appears in the sidebar

---

## 2. Case List Tabs (assignee filtering)

The case list now has **4 tabs** at the bottom:

| Tab | Filter |
|-----|--------|
| **All cases** | No filter |
| **My cases** | Assigned to current user |
| **Team cases** | Assigned to any of the current user's teams |
| **Unassigned cases** | No user or team assigned |

### What to test:
- Tab switching filters cases correctly
- Each tab shows a different no-results message when empty
- Switching tabs while rows are selected shows a confirmation modal
- Tab visibility can be configured via `config.visibleCaseListTabs`

---

## 3. Team + User Assignment on Cases

Cases can now be assigned to **both** a user and a team (independently).

- **API**: `POST /api/v1/document/{documentId}/assign` with `assigneeId` and/or `assignedTeamKey`
- **UI**: The assignment component now has **two dropdowns** — "Select User" and "Select Team"
- Unassign removes both user and team

### What to test:
- Assign a team to a case, assign a user to a case, assign both
- Unassign all — removes both
- Assignment pills show correct icons (group icon for team, user icon for user)
- Click-to-edit flow on the assignment pills

---

## 4. Team + User Assignment on Tasks

Same dual-assignment model applies to tasks.

- **API**: `POST /api/v1/task/{taskId}/assign` with `assignee` and/or `assignedTeamKey`
- **API**: `GET /api/v1/task/{taskId}/candidate-team` — available teams
- **UI**: Task detail modal shows the same dual-assignment component

### What to test:
- Assign/unassign team and user on a task from the task detail modal
- Candidate teams list loads correctly
- Task list shows team assignment info

---

## 5. Permissions (PBAC)

### New Team resource permissions

| Action | Description | Default: ADMIN | Default: USER |
|--------|-------------|:-:|:-:|
| `view` | View a team | Yes | Only if member |
| `view_list` | List teams | Yes | Only if member |
| `create` | Create team | Yes | No |
| `modify` | Edit team | Yes | Only if member |
| `delete` | Delete team | Yes | Only if member |
| `assign` | Add/remove members | Yes | Only if member |
| `assignable` | Team appears in assignment dropdowns | Yes | Yes (global) |

### Document permissions (team-based)

- Users can **view, view_list, modify, export** cases assigned to their team (condition: `assignedTeamKey` in `${currentUserTeams}`)
- Users can **claim** cases that are assigned to their team but have no individual assignee (conditions: `assigneeId` is null AND `assignedTeamKey` in `${currentUserTeams}`)

### What to test:
- **As admin**: full CRUD on teams, can assign teams to cases/tasks
- **As regular user (team member)**: can see and manage teams they belong to, can see cases assigned to their team via the "Team cases" tab, can claim unassigned team cases
- **As regular user (not in any team)**: "Team cases" tab shows no results, cannot see/manage teams
- Remove a permission and verify the corresponding UI element (button/tab/dropdown) disappears

---

## 6. Team Lifecycle Events

When a team is updated or deleted, changes cascade:

- **Team renamed** → title updates on all assigned cases and tasks
- **Team deleted** → team assignment is cleared from all cases and tasks, `TaskTeam` records are removed

### What to test:
- Rename a team and verify the new name appears on previously assigned cases/tasks
- Delete a team and verify cases/tasks no longer show the team assignment

---

## 7. Other Changes

- **Overflow menu component** — new reusable overflow menu (used in team detail header, also in case management detail and other places)
- **Case list refactor** — the case list was refactored to use an orchestration service (`CaseListOrchestrationService`) for better state management
- **Import/export** — teams can be imported/exported via the auto-deployment system (`default-team.team.json`)
- **Document table value resolver** — `doc:assignedTeamTitle` is now available for use in case list column configurations

documentation/release-notes/13.x.x/13.20.0/README.md
```diff
@@ -27,6 +27,21 @@
 * Fixed a bug where task auto-assignment ignored permission restrictions. When auto-assign was enabled, the case assignee was automatically assigned to tasks regardless of whether their permissions allowed it.
 * Cleaned up unused code for task notifications, solving an error about `email_notification_settings_days` that appeared
   once a day.
+
+* **Secure /users endpoint with access control**
+
+  A new access control resource type has been added `com.ritense.valtimo.contract.authentication.User`.
```
ivo-ritense: Is this also a breaking change? Are normal candidates still available as before?

ivo-ritense:
The way that the frontend creates the key, but this is not visible or controllable is weird. Lets say you create a team called "Team" and later rename that to some other name. If you then create a new team called "Team" it suddenly edits the name of the previous team, without the user understanding why that happened.

ivo-ritense:
Teams with capitals in the name are not allowed?

ivo-ritense:
@peetritense what do you think?

UI feedback: i would prefer it if when creating a team the name field was automatically selected, and when hitting enter it would submit the modal

ivo-ritense:
Do we want team to be listed alphabetically? Right now it looks like it's based on time of creation, which makes it hard to find the one you're looking for.

> Edit Klaas: Changed, it now sorting alphabetically

ivo-ritense:
There's no highlight when hovering over the context menu button, making it unclear when you're clicking the row and when you're clicking the context menu button.

peetritense:
> @peetritense what do you think?
> 
> UI feedback: i would prefer it if when creating a team the name field was automatically selected, and when hitting enter it would submit the modal

In this case it's a little difficult. We have multiple places in the product where only 1 input is present in a modal. The carbon guidelines say the focus should be on the first input, and then tabbing to a the first primary button. The input field should be selected when opening the modal though, and that doesn't happen now.

I would however like to see the 'new' design implemented, so it works the same as adding a member to a team. I think that solves the issue as well.

ivo-ritense:
Automatically assigning user tasks only assigns to the user, not to the team. is this intended?

ivo-ritense:
After update of a team name, the name is not reflected in cases where the team has been assigned.
image
image

Same goes for task.

ivo-ritense:
Assigning cases from multi select in the case list only allows you to assign to a user, not to a team. Is this intended?

ivo-ritense:
Unclear team name requirements in general:

image

ivo-ritense:
Cut of of length of team name seems to work, but on case screen it still makes the assignee name a bit weird: 
image

Same goes for task:
image

ivo-ritense:
Do we have no team tab on the task list?

ivo-ritense:
How do i export and import teams? Or is this only on startup?

> Edit Klaas: The TeamImporter is only for autodeployment right now. Its not yet possible to export even though I made a TeamExporter. Im kinda making these for the Deployer -feature

ivo-ritense:
"As regular user (team member): can see and manage teams they belong to, can see cases assigned to their team via the "Team cases" tab, can claim unassigned team cases"

I can see the team in the list and go to the detail page. Through the UI i cannot:
- Assign people
- Remove people
- Edit the team
- Delete the team


Even though i am in the team and have this permission:

```
    {
        "resourceType": "com.ritense.team.domain.Team",
        "actions": [
            "assign",
            "view",
            "view_list",
            "modify",
            "delete"
        ],
        "conditions": [
            {
                "type": "field",
                "field": "users",
                "operator": "list_contains",
                "value": "${currentUsername}"
            }
        ]
    }
```


Klaas: The problem is that the frontend is not sending the permission context for the assign, view, modify, delete.

ivo-ritense:
Should we even show the teams page when the user doesn't have view_list permission on teams?

ivo-ritense:
When deleting team, unassigning seems to work for tasks, but not for cases.

Edit Klaas: Fixed

ivo-ritense:
Merge of the configuration warnings went wrong (already known) where the "Set as active version" button is available even when we know there are configuration issues.

backend/case/src/main/java/com/ritense/document/web/rest/DocumentResource.java
Klaas-Ritense: Fixed

backend/core/src/main/resources/config/liquibase/initial-setup/task/changelog/20260317-create-table-task-team.xml
Klaas-Ritense: Because it performs better. Because joining the team table is not possible here. This task_team table is in the core module, while the team table is in the team module

backend/team/src/main/kotlin/com/ritense/team/domain/TeamUser.kt
Klaas-Ritense: Fixed

backend/team/src/main/kotlin/com/ritense/team/service/TeamManagementServiceImpl.kt
Klaas-Ritense: Changed

backend/team/src/main/kotlin/com/ritense/team/service/TeamManagementServiceImpl.kt
Klaas-Ritense: Changed

backend/team/src/main/kotlin/com/ritense/team/web/rest/TeamResource.kt
Klaas-Ritense: Changed to error(..)

backend/team/src/main/kotlin/com/ritense/team/web/rest/TeamResource.kt
Klaas-Ritense: Removed

documentation/release-notes/13.x.x/13.20.0/README.md
Klaas-Ritense: Removed. This release-note was part of another PR and is already present in those release-notes

backend/team/src/main/kotlin/com/ritense/team/web/rest/TeamResource.kt
Klaas-Ritense: It does not perform well. The only solution I see is to make a paginated dropdown

mbritense:
> Should we even show the teams page when the user doesn't have view_list permission on teams?

We decided that we won't built an automatic hide, since the environment 

> The way that the frontend creates the key, but this is not visible or controllable is weird. Lets say you create a team called "Team" and later rename that to some other name. If you then create a new team called "Team" it suddenly edits the name of the previous team, without the user understanding why that happened.

Should be fixed, now uses auto key component

mbritense:
> Teams with capitals in the name are not allowed?

I think so yes? Now I think this is fixed with the auto key component

mbritense:
> "As regular user (team member): can see and manage teams they belong to, can see cases assigned to their team via the "Team cases" tab, can claim unassigned team cases"
> 
> I can see the team in the list and go to the detail page. Through the UI i cannot:
> 
> * Assign people
> * Remove people
> * Edit the team
> * Delete the team
> 
> Even though i am in the team and have this permission:
> 
> ```
>     {
>         "resourceType": "com.ritense.team.domain.Team",
>         "actions": [
>             "assign",
>             "view",
>             "view_list",
>             "modify",
>             "delete"
>         ],
>         "conditions": [
>             {
>                 "type": "field",
>                 "field": "users",
>                 "operator": "list_contains",
>                 "value": "${currentUsername}"
>             }
>         ]
>     }
> ```
> 
> Klaas: The problem is that the frontend is not sending the permission context for the assign, view, modify, delete.

Should be fixed, now sending context

mbritense:
> Merge of the configuration warnings went wrong (already known) where the "Set as active version" button is available even when we know there are configuration issues.

Should be fixed

mbritense:
> Cut of of length of team name seems to work, but on case screen it still makes the assignee name a bit weird:

should be fixed

mbritense:
> After update of a team name, the name is not reflected in cases where the team has been assigned. image image
> 
> Same goes for task.

should be fixed for cases

mbritense (approved):

PR #435, opened 2026-03-10
Added permission checks to ensure users who cannot be assigned to a task aren't assigned to that task.

marijnritense:
Solves https://github.com/generiekzaakafhandelcomponent/gzac-issues/issues/428

PR #446, opened 2026-03-12
609 Secure /users endpoints with PBAC

Klaas-Ritense:
https://github.com/generiekzaakafhandelcomponent/gzac-issues/issues/609

### DIFF
--- a/backend/keycloak-iam/src/main/java/com/valtimo/keycloak/service/KeycloakUserManagementService.java
+++ b/backend/keycloak-iam/src/main/java/com/valtimo/keycloak/service/KeycloakUserManagementService.java
@@ -16,13 +16,13 @@
 
 package com.valtimo.keycloak.service;
 
-import static com.ritense.valtimo.contract.Constants.SYSTEM_ACCOUNT;
-import static java.util.Comparator.comparing;
-import static java.util.Comparator.naturalOrder;
-import static java.util.Comparator.nullsLast;
-
+import com.ritense.authorization.Action;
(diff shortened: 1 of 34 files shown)


backend/keycloak-iam/src/main/java/com/valtimo/keycloak/service/KeycloakUserManagementService.java
```diff
@@ -124,12 +136,21 @@ public List<ManageableUser> getAllUsers() {
...
-        throw new NotImplementedException();
+        List<ManageableUser> users = getAllUsers();
+        return PageableExecutionUtils.getPage(users, pageable, users::size);
     }
 
     @Override
     public Page<ManageableUser> queryUsers(String searchTerm, Pageable pageable) {
-        throw new NotImplementedException();
+        String normalizedSearchTerm = searchTerm.toLowerCase().trim();
+        List<ManageableUser> users = getAllUsers().stream()
```
marijnritense: The `/users` endpoint from key cloak allows searching the username, first name, lastname as well as email address, as is documented [here](https://github.com/valtimo-platform/valtimo/pull/435), so that way we don't have to retrieve all users and filter them ourselves.

Klaas-Ritense: changed

marijnritense:
Tested existing functionality in FE as well as used the existing users endpoint with searchterm to test that functionality. LGTM

marijnritense (approved):

PR #480, opened 2026-03-20
434 case definition pbac containers

Klaas-Ritense:
https://github.com/generiekzaakafhandelcomponent/gzac-issues/issues/434

## New container relationships

| Source resource type           | Container resource type        | Description                                                    |
|--------------------------------|--------------------------------|----------------------------------------------------------------|
| `JsonSchemaDocument`           | `CaseDefinition`               | Scope document permissions to a case definition                |
| `JsonSchemaDocumentDefinition` | `CaseDefinition`               | Scope document definition permissions to a case definition     |
| `OperatonExecution`            | `CaseDefinition`               | Scope process instance permissions to a case definition        |
| `OperatonExecution`            | `JsonSchemaDocument`           | Scope process instance permissions to document properties      |
| `OperatonProcessDefinition`            | `CaseDefinition`           | Scope process definition permissions to case definition      |


backend/building-block/src/main/kotlin/com/ritense/buildingblock/repository/JsonSchemaDocumentCaseDefinitionMapper.kt
```diff
@@ -0,0 +1,104 @@
...
+import com.ritense.valtimo.contract.document.CaseDocumentResolver
+import jakarta.persistence.criteria.AbstractQuery
+import jakarta.persistence.criteria.CriteriaBuilder
+import jakarta.persistence.criteria.Root
+import java.util.UUID
+import org.springframework.stereotype.Component
+
+@Component
+@SkipComponentScan
+class JsonSchemaDocumentCaseDefinitionMapper(
```
ivo-ritense, line 42: I'm not sure about putting this in the building block module. I understand why, but this also means that this link is not available without the building block module. i'll leave this to you to see if this should change.

backend/building-block/src/test/kotlin/com/ritense/buildingblock/repository/JsonSchemaDocumentCaseDefinitionMapperIntegrationTest.kt
```diff
@@ -0,0 +1,160 @@
...
+    private val mapper: JsonSchemaDocumentCaseDefinitionMapper,
+    private val buildingBlockInstanceRepository: BuildingBlockInstanceRepository,
+    private val jsonSchemaDocumentRepository: JsonSchemaDocumentRepository,
+    private val buildingBlockDefinitionDeploymentService: BuildingBlockDefinitionDeploymentService,
+    private val caseDefinitionDeploymentService: CaseDefinitionDeploymentService,
+) : BaseIntegrationTest() {
+
+    @Test
+    fun `mapQuery should correctly link JsonSchemaDocument to CaseDefinition for building block blueprint`() {
+        buildingBlockDefinitionDeploymentService?.deployOnStartup()
```
ivo-ritense: Not sure why these are needed

backend/building-block/src/test/kotlin/com/ritense/buildingblock/repository/JsonSchemaDocumentCaseDefinitionMapperIntegrationTest.kt
```diff
@@ -0,0 +1,160 @@
...
+            sub.where(*predicates)
+
+            result.joinPredicate
+        }
+
+        val foundDocuments = jsonSchemaDocumentRepository.findAll(spec)
+        assertThat(foundDocuments).isEmpty()
+    }
+
+    private fun createDocument(definitionId: JsonSchemaDocumentDefinitionId): JsonSchemaDocument {
```
ivo-ritense: Why is this done is such a complicated way?

backend/process-document/src/main/kotlin/com/ritense/processdocument/repository/OperatonExecutionCaseDefinitionMapper.kt
```diff
@@ -0,0 +1,111 @@
...
+class OperatonExecutionCaseDefinitionMapper(
+    private val processDefinitionCaseDefinitionRepository: ProcessDefinitionCaseDefinitionRepository,
+    private val caseDefinitionService: CaseDefinitionService,
+    private val executionDocumentMapper: OperatonExecutionJsonSchemaDocumentMapper,
+    private val authorizationService: AuthorizationService,
+    private val queryDialectHelper: QueryDialectHelper,
+) : AuthorizationEntityMapper<OperatonExecution, CaseDefinition> {
+
+    override fun mapRelated(entity: OperatonExecution): List<CaseDefinition> {
+        // Path A: via ProcessDefinitionCaseDefinition
```
ivo-ritense, line 51: I don't think path A is reliable. what happens when a document definition is not longer linked because it was replaced by a newer version?

On the other hand, is there a reason not to always use path B? i don't see it right now.

backend/process-document/src/main/kotlin/com/ritense/processdocument/repository/OperatonExecutionCaseDefinitionMapper.kt
```diff
@@ -0,0 +1,111 @@
...
+        cb: CriteriaBuilder
+    ): AuthorizationEntityMapperResult<CaseDefinition> {
+
+        val sub = query.subquery(Int::class.java)
+        val cd = sub.from(CaseDefinition::class.java)
+        val cdId = cd.get<Any>("id")
+
+        val pd = root.join<OperatonExecution, OperatonProcessDefinition>("processDefinition", JoinType.LEFT)
+
+        // Path A: via ProcessDefinitionCaseDefinition (cross-join in same subquery)
```
ivo-ritense, line 78: See above

backend/process-document/src/main/kotlin/com/ritense/processdocument/repository/OperatonExecutionJsonSchemaDocumentMapper.kt
```diff
@@ -0,0 +1,82 @@
...
+import org.springframework.stereotype.Component
+
+@Component
+@SkipComponentScan
+class OperatonExecutionJsonSchemaDocumentMapper(
+    private val processDocumentService: OperatonProcessJsonSchemaDocumentService,
+    private val queryDialectHelper: QueryDialectHelper,
+) : AuthorizationEntityMapper<OperatonExecution, JsonSchemaDocument> {
+
+    override fun mapRelated(entity: OperatonExecution): List<JsonSchemaDocument> {
```
ivo-ritense, line 40: What is the thought behind not supporting building blocks here? That is assuming that we want to relate a process instance to a case, not a document. Not sure what makes sense.

backend/process-document/src/main/kotlin/com/ritense/processdocument/repository/OperatonExecutionCaseDefinitionMapper.kt
Klaas-Ritense: Path A is not always reliable. For example: The verzoek plugin starts a process that is not linked in the processDefinitionCaseDefinition table.

Path B is a fallback right now for when path A fails.

I could remove path A. But then again it does perform better:

 ```
 Question: is there a performace difference between path A and B                                                                                                                                            

⏺ Yes, significant difference in mapQuery:

  Path A joins through ProcessDefinitionCaseDefinition — a small lookup table. The join is on processDefinition.id which is already available from the execution's LEFT JOIN. Very cheap.                    
   
  Path B cross-joins JsonSchemaDocument into the subquery and matches execution.businessKey against CAST(doc.id AS VARCHAR). This:                                                                           
  - Brings a potentially large table (json_schema_document) into the subquery                                                                                                                              
  - Requires a string cast + comparison (UUID → VARCHAR) which can't use indexes efficiently                                                                                                                 
  - When building-block is present, adds a nested EXISTS subquery against building_block_instance                                                                                                          
                                                                                                                                                                                                             
  For mapRelated, Path A is a single query on a small table vs Path B loading the full document entity.                                                                                                      
                                                                                                                                                                                                             
  So Path A is a meaningful performance optimization when a ProcessDefinitionCaseDefinition link exists.
```

backend/process-document/src/main/kotlin/com/ritense/processdocument/repository/OperatonExecutionCaseDefinitionMapper.kt
Klaas-Ritense: See above

backend/process-document/src/main/kotlin/com/ritense/processdocument/repository/OperatonExecutionJsonSchemaDocumentMapper.kt
Klaas-Ritense: Ive added it now

backend/building-block/src/main/kotlin/com/ritense/buildingblock/repository/JsonSchemaDocumentCaseDefinitionMapper.kt
Klaas-Ritense: Ive added a second mapper to the case module for when the building-block module is not included

backend/building-block/src/test/kotlin/com/ritense/buildingblock/repository/JsonSchemaDocumentCaseDefinitionMapperIntegrationTest.kt
Klaas-Ritense: Removed

backend/building-block/src/test/kotlin/com/ritense/buildingblock/repository/JsonSchemaDocumentCaseDefinitionMapperIntegrationTest.kt
Klaas-Ritense: Changed

ivo-ritense:
The link between operaton execution and case definition doesn't work completely. It is possible to give people permission this way to start supporting processes based on the case definition. What does not work however is filtering the list of available supporting processes. It will always show the whole list regardless of the permission set up. When actually clicking the process to start it, you still get an error saying you cannot do it. The list should be filtered in an expected way based on the processes someone can start.

Please test this for:
- execution -> case definition
-  execution -> jsonschemadocument 
-  execution -> jsonschemadocument -> case definition

Klaas-Ritense:
@ivo-ritense Most mappings should now work. 

This mapping is problematic for action 'create': execution -> jsonschemadocument
Because there is not always a document available

ivo-ritense (approved):

PR #485, opened 2026-03-24
Added patch for datagrid

ivo-ritense:


ivo-ritense:
We're releasing this as 12.28.1 after all

PR #507, opened 2026-03-30
story: redact sensitive headers

mbritense:
closes https://github.com/orgs/generiekzaakafhandelcomponent/projects/5/views/3?sliceBy%5Bvalue%5D=WEEK+14&pane=issue&itemId=170455372&issue=generiekzaakafhandelcomponent%7Cgzac-issues%7C657

### DIFF
--- a/backend/web/src/main/kotlin/com/ritense/valtimo/web/logging/LoggingRestClientCustomizer.kt
+++ b/backend/web/src/main/kotlin/com/ritense/valtimo/web/logging/LoggingRestClientCustomizer.kt
@@ -1,5 +1,5 @@
 /*
- * Copyright 2015-2024 Ritense BV, the Netherlands.
+ * Copyright 2015-2026 Ritense BV, the Netherlands.
  *
  * Licensed under EUPL, Version 1.2 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


sofiaIvarsRitense (approved):

ThomasMinkeRitense:
@mbritense @theo-ritense I think the issue here is mostly the request/response body. For the request headers, we're only logging the keys at the moment, which I think is fine. I've never seen sensitive data in the response headers, so I'm wondering if the headers are a problem at all.

backend/web/src/main/kotlin/com/ritense/valtimo/web/logging/LoggingRestClientCustomizer.kt
```diff
@@ -142,5 +153,18 @@ class LoggingRestClientCustomizer : RestClientCustomizer, ClientHttpRequestInter
 
     companion object {
         val logger = KotlinLogging.logger {}
+
+        private const val REDACTED = "REDACTED"
+
+        internal val SENSITIVE_HEADERS = setOf(
```
theo-ritense, line 159: We should not do blacklisting, but whitelisting. You never know what custom headers a server might send, and what's in there. Please whitelist known (standard) non-sensitive headers instead.

backend/web/src/main/kotlin/com/ritense/valtimo/web/logging/LoggingRestClientCustomizer.kt
```diff
@@ -45,9 +46,9 @@ class LoggingRestClientCustomizer : RestClientCustomizer, ClientHttpRequestInter
         val response = execution.execute(request, requestBody)
         val requestBodyHead = requestBody.take(DEFAULT_BUFFER_SIZE).toByteArray()
```
theo-ritense, line 47: Only taking the `DEFAULT_BUFFER_SIZE` does not guarantee there's not sensitive data in that part.

PR #549, opened 2026-04-10
locked FE dependencies to specific versions

marijnritense:
Solves https://github.com/generiekzaakafhandelcomponent/atlas-internal/issues/391


Klaas-Ritense (approved):

Klaas-Ritense:
Failed to build frontend.
Reproduction steps:
1. npm run clean
2. npm i
3. npm run libs-build-all
4. npm start
5. Notice error: `/angular-split.mjs:211:21-27 - Error: export 'output' (imported as 'output') was not found in '@angular/core' `

Claude said:
> The root cause: when locking versions, @angular/core was pinned to 17.2.2 (the caret baseline), but the lock file previously resolved it to 17.3.12. The public output() API — used by angular-split@18.0.0-beta.1 — was added in Angular 17.3, so 17.2.2 doesn't export it.

PR #556, opened 2026-04-14
feat: improve devMode

mbritense:
- DevMode no longer crashes when you edit multiple libraries at the same time. Previously, changing files in several libraries in quick succession would cause build failures and break the dev server, forcing a restart. Now it handles this gracefully.
- Libraries are built in the correct order automatically. The build system now reads the dependency tree from package.json at startup and makes sure a library is never built before the libraries it depends on are finished. Before, builds ran independently with no awareness of each other.
- The dev server no longer shows "Cannot find module" errors during rebuilds. Previously, rebuilding a library would briefly delete its output folder, causing the dev server to fail and sometimes not recover. Now the build happens in a temporary folder and gets swapped in atomically, so the dev server never sees a broken state.
- One coordinated process instead of many independent ones. The old approach spawned a separate watcher and rebuilder for each library, all competing with each other. Now a single process coordinates everything with a queue, avoiding race conditions.
- Rapid edits are batched together. If you save a file multiple times quickly, or edit several files across libraries in a short burst, the system debounces and groups the work instead of kicking off redundant builds.
- The dev server recompiles exactly once after a batch of rebuilds, not once per library. This makes the feedback loop faster when multiple libraries change at the same time.

### DIFF
--- a/frontend/scripts/dev-mode.js
+++ b/frontend/scripts/dev-mode.js
@@ -1,5 +1,5 @@
 /*
- * Copyright 2015-2024 Ritense BV, the Netherlands.
+ * Copyright 2015-2026 Ritense BV, the Netherlands.
  *
  * Licensed under EUPL, Version 1.2 (the "License");
  * you may not use this file except in compliance with the License.
@@ -213,7 +213,9 @@ function buildLib(libName, {safeSwap = false} = {}) {
           if (!fs.existsSync(realDest) && fs.existsSync(backupDest)) {
             fs.renameSync(backupDest, realDest);
           }
-        } catch { /* best effort */ }
+        } catch {
+          /* best effort */
+        }
       }
     }
 
--- a/frontend/scripts/dev-mode.js
+++ b/frontend/scripts/dev-mode.js
@@ -1,23 +1,111 @@
-const {spawn, spawnSync} = require('child_process');
+/*
+ * Copyright 2015-2024 Ritense BV, the Netherlands.
+ *
+ * Licensed under EUPL, Version 1.2 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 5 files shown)


frontend/scripts/dev-mode.js
```diff
@@ -1,23 +1,111 @@
-const {spawn, spawnSync} = require('child_process');
+/*
+ * Copyright 2015-2024 Ritense BV, the Netherlands.
```
sofiaIvarsRitense: 2026

frontend/scripts/dev-mode.js
```diff
@@ -1,23 +1,111 @@
...
+
+console.log(
+  `\n📦 Loaded dependency graph: ${allLibs.length} libraries across ${sortedLevelNumbers.length} levels\n`
+);
+for (const levelNum of sortedLevelNumbers) {
+  const libs = levels.get(levelNum);
+  const mode = levelBuildModes.get(levelNum);
+  console.log(`  Level ${levelNum} (${mode}): ${libs.join(', ')}`);
+}
+console.log('');
```
sofiaIvarsRitense, line 61: Is this console.log necessary?

frontend/scripts/rebuild-lib.js
```diff
@@ -1,35 +1,74 @@
+/*
+ * Copyright 2015-2024 Ritense BV, the Netherlands.
```
sofiaIvarsRitense: 2026

frontend/scripts/lib-dependency-graph.js
```diff
@@ -0,0 +1,311 @@
+/*
+ * Copyright 2015-2024 Ritense BV, the Netherlands.
```
sofiaIvarsRitense: 2026

frontend/scripts/dev-mode.js
mbritense: Yes it's useful information

sofiaIvarsRitense (approved):

PR #557, opened 2026-04-14
Pick plugin configurations on import

### LINKED ISSUE #656: Configure plugins after importing
## Why
*Why do we need this feature?*

When importing a case definition, the plugin configurations might be different on the target environment. This can be because you forgot to add a certain plugin configuration, or you received the case definition from someone else, so that uuid don't match.

Currently when importing it sometimes fails without being clear what's wrong. It's easier to have the import succeed and after that fix any resulting issues.

## How
*How do we envision this feature working?*

- You start the import as normal
- After picking the case definition name/key a new step is added.
  - On this step you pick the plugin configurations in the target system, based on the plugins used in the source system.
  - The plugin configurations are prefilled with the ones that we know already (uuid matches)
  - These fields are not required, and a message is displayed telling the user this.
- After completing the import, there might still be process links that were not configured, for this a new tab is added to the case definition screen.
- This allows the user to set the plugin configurations when they can. They also have time to create the missing configurations.

## What
*What needs to be done to implement this feature?*

- **Ideas**: 
  - Case definition should not be startable (or be made active) before the issues are resolved
  - Configuration issue framework can be re-used.

- **Other**: 
  - Design to be made @peetritense 

- **Challenges**:
  -  Plugin process links should have the plugin configuration be made optional in the code, but not when a user configures them
  - This also impacts the building block process links.
  - The original plugin information needs to be added in the export, and saved as meta data for at least until the configuration is fixed.

## Additional Context
- **Stakeholders**: Requested by Frits and Mark from Team?

---

ivo-ritense:
Closes: https://github.com/generiekzaakafhandelcomponent/gzac-issues/issues/656


Klaas-Ritense:
@ivo-ritense Please add release notes and documenation. Also look at the coderabbitai comments, please

backend/case/src/main/kotlin/com/ritense/case/service/CaseDefinitionImportPreviewService.kt
```diff
@@ -18,41 +18,124 @@ package com.ritense.case.service
...
+                objectMapper.readTree(entry.content.toString(Charsets.UTF_8))
+            } catch (_: Exception) {
+                continue
+            }
+
+            if (jsonTree !is ArrayNode) continue
+
+            for (node in jsonTree) {
+                val processLinkType = node.path("processLinkType").asText(null) ?: continue
+                if (processLinkType != "plugin") continue
```
Klaas-Ritense: Lots of process-link and plugin-configuration domain knowledge inside the core module feels wrong

backend/core/src/main/resources/config/liquibase/13-24-0/13-24-0-master.xml
Klaas-Ritense, line 1: 13.24.0 has already been released

backend/process-link/src/main/kotlin/com/ritense/processlink/service/ProcessLinkService.kt
```diff
@@ -140,15 +145,21 @@ class ProcessLinkService(
                 // Hibernate does not allow 2 different objects with the same identifier in the session, so do delete + insert instead
                 processLinkRepository.delete(processLinkToUpdate)
             }
-            processLinkRepository.save(processLinkUpdated)
+            val saved = processLinkRepository.save(processLinkUpdated)
+        applicationEventPublisher.publishEvent(ProcessLinkChangedEvent(updateRequest./clprocessDefinitionId))
```
Klaas-Ritense: Build is failing

backend/case/src/main/kotlin/com/ritense/case/service/CaseDefinitionImportPreviewService.kt
ivo-ritense: Agree, moved this to the correct module.

backend/core/src/main/resources/config/liquibase/13-24-0/13-24-0-master.xml
ivo-ritense: Changed to 13.25.0

marijnritense (approved):

frontend/projects/valtimo/case-management/src/lib/components/case-management-detail/tabs/case-management-general/components/case-management-missing-plugin-configurations/case-management-missing-plugin-configurations.component.html
```diff
@@ -0,0 +1,74 @@
...
+  ~
+  ~ Unless required by applicable law or agreed to in writing, software
+  ~ distributed under the License is distributed on an "AS IS" basis,
+  ~ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  ~ See the License for the specific language governing permissions and
+  ~ limitations under the License.
+  -->
+
+@if (hasIssue$ | async) {
+  @if (mappingRows$ | async; as rows) {
```
sofiaIvarsRitense: The 2 previous if can be combined:
`@if ((hasIssue$ | async) && (mappingRows$ | async); as rows) {`

Or even moved to the typescript
`public readonly visibleRows$ = combineLatest([
  this.hasIssue$,
  this.mappingRows$,
]).pipe(
  map(([hasIssue, rows]) => (hasIssue && rows?.length ? rows : null)),
);`

`@if (visibleRows$ | async; as rows) {`

frontend/projects/valtimo/case-management/src/lib/components/case-management-detail/tabs/case-management-general/components/case-management-missing-plugin-configurations/case-management-missing-plugin-configurations.component.ts
```diff
@@ -0,0 +1,264 @@
...
+  getCaseManagementRouteParams,
+  GlobalNotificationService,
+} from '@valtimo/shared';
+import {BehaviorSubject, filter, Observable, switchMap, take} from 'rxjs';
+import {CaseManagementService} from '../../../../../../services';
+import {DanglingPluginConfiguration} from '../../../../../../models/case-deployment.model';
+
+type PluginMappingStatus = 'available' | 'no-configurations' | 'not-installed';
+
+interface MappingRow {
```
sofiaIvarsRitense: Can you move this interface (MappingRow) and this type (PluginMappingStatus) to the corresponding model file, please?

sofiaIvarsRitense (approved):

PR #565, opened 2026-04-15
story: building block dmn support

mbritense:
closes https://github.com/orgs/generiekzaakafhandelcomponent/projects/5/views/3?sliceBy%5Bvalue%5D=WEEK+16&pane=issue&itemId=160405842&issue=generiekzaakafhandelcomponent%7Catlas-internal%7C231


sofiaIvarsRitense (approved):

Klaas-Ritense:
I noticed many duplicate files and asked Claude about it. 



**The duplication is real**

For almost every artifact type (form, process, decision, document definition, form-flow, artwork, process-link…) there are parallel `BuildingBlockXImporter` / `BuildingBlockXExporter` classes that mirror their existing counterparts. Comparing `BuildingBlockFormDefinitionExporter` against `FormDefinitionExporter`, the only real differences are:

1. Which ID scopes the artifact (`caseDefinitionId` → `buildingBlockDefinitionId`)
2. Which service it delegates to
3. The path prefix (`config/case/…` vs `config/building-block/…`)
4. The importer `type()` / `dependsOn()` tokens (`FORM`/`CASE_DEFINITION` vs `BUILDING_BLOCK_FORM_DEFINITION`/`BUILDING_BLOCK_DEFINITION`)

Everything else — regex, parsing, transactional semantics, entity layer — is identical.

**Is there a valid case for duplicating?**

**Sometimes, but not really here.** Legitimate reasons to duplicate rather than abstract include:

- The two things look similar _today_ but are expected to diverge soon, and a premature abstraction would lock in the wrong seams
- The two things live in different bounded contexts with genuinely independent lifecycles and ownership
- The shared code would require leaky generics or conditionals everywhere that obscure intent more than the duplication

None of those really apply here. `BuildingBlockFormDefinitionService` at `backend/building-block/.../BuildingBlockFormDefinitionService.kt:36` uses the **same** `FormDefinitionRepository` and the **same** `FormIoFormDefinition` entity as the case-scoped service — it just filters on `BlueprintType.BUILDING_BLOCK` instead of `BlueprintType.CASE`. The persistence layer already generalized the concept (via the `BlueprintType` discriminator on `FormDefinitionBlueprintId`). The service/importer/exporter layer then _re-specializes_ it, which is the opposite of what you'd want.

**How it could have been done differently**

A few options, roughly from least to most invasive:

1. **Parameterize scope on the existing importer/exporter.** Replace `CaseDefinitionId` in the request with a sealed `OwnerScope` (either `CaseScope(caseDefinitionId)` or `BuildingBlockScope(buildingBlockDefinitionId)`). The exporter computes the path prefix from the scope; the importer dispatches on a prefix in the regex. One class, two flavors.
2. **A generic `BlueprintScopedExporter<T : BlueprintId>` + registry.** The exporter core logic is identical; only the path template and the ID lookup vary. Inject a `PathTemplate` and a `ScopedDefinitionLookup<T>` and you get one implementation per artifact type instead of two.
3. **Treat "owner" as a first-class concept end-to-end.** Since the repository already keys on `(BlueprintType, key, versionTag)`, push `BlueprintType` all the way up. Importers then key off `type()` returning something like `FORM_DEFINITION` universally, with the dependency graph expressed via scope rather than via a separate `BUILDING_BLOCK_*` import-type enum. The `ValtimoImportTypes` doubling (CASE_DEFINITION / FORM vs BUILDING_BLOCK_DEFINITION / BUILDING_BLOCK_FORM_DEFINITION) is the symptom that the scope wasn't modeled as data.
4. **Strategy for path layout only.** If the divergence is truly just "where on disk does the config bundle live," inject a `BundleLayout` strategy and keep a single importer/exporter. This is the minimum-viable refactor.

**Why it probably happened**

My guess: building-block was added after case-scoped deployment was already working, under time pressure, and copy-paste-modify is the lowest-risk way to ship without destabilizing the existing case path. That's a legitimate _tactical_ choice — but it leaves a structural debt, because now every future change to import/export mechanics has to be made in two places, and the `ValtimoImportTypes` enum keeps accreting parallel tokens. Option 1 or 4 above would be a reasonable consolidation pass once the feature is stable.

Short version: there isn't a strong design reason for the duplication, just a historical one. The persistence layer already proves the right abstraction exists; the upper layers just didn't follow through.

Klaas-Ritense:
@mbritense 
- Missing documentation
- Missing release notes
- Lots of open coderabbit comments
- linting issues
- Failing to create test environment

sofiaIvarsRitense (approved):

mbritense:
> I noticed many duplicate files and asked Claude about it.
> 
> **The duplication is real**
> 
> For almost every artifact type (form, process, decision, document definition, form-flow, artwork, process-link…) there are parallel `BuildingBlockXImporter` / `BuildingBlockXExporter` classes that mirror their existing counterparts. Comparing `BuildingBlockFormDefinitionExporter` against `FormDefinitionExporter`, the only real differences are:
> 
> 1. Which ID scopes the artifact (`caseDefinitionId` → `buildingBlockDefinitionId`)
> 2. Which service it delegates to
> 3. The path prefix (`config/case/…` vs `config/building-block/…`)
> 4. The importer `type()` / `dependsOn()` tokens (`FORM`/`CASE_DEFINITION` vs `BUILDING_BLOCK_FORM_DEFINITION`/`BUILDING_BLOCK_DEFINITION`)
> 
> Everything else — regex, parsing, transactional semantics, entity layer — is identical.
> 
> **Is there a valid case for duplicating?**
> 
> **Sometimes, but not really here.** Legitimate reasons to duplicate rather than abstract include:
> 
> * The two things look similar _today_ but are expected to diverge soon, and a premature abstraction would lock in the wrong seams
> * The two things live in different bounded contexts with genuinely independent lifecycles and ownership
> * The shared code would require leaky generics or conditionals everywhere that obscure intent more than the duplication
> 
> None of those really apply here. `BuildingBlockFormDefinitionService` at `backend/building-block/.../BuildingBlockFormDefinitionService.kt:36` uses the **same** `FormDefinitionRepository` and the **same** `FormIoFormDefinition` entity as the case-scoped service — it just filters on `BlueprintType.BUILDING_BLOCK` instead of `BlueprintType.CASE`. The persistence layer already generalized the concept (via the `BlueprintType` discriminator on `FormDefinitionBlueprintId`). The service/importer/exporter layer then _re-specializes_ it, which is the opposite of what you'd want.
> 
> **How it could have been done differently**
> 
> A few options, roughly from least to most invasive:
> 
> 1. **Parameterize scope on the existing importer/exporter.** Replace `CaseDefinitionId` in the request with a sealed `OwnerScope` (either `CaseScope(caseDefinitionId)` or `BuildingBlockScope(buildingBlockDefinitionId)`). The exporter computes the path prefix from the scope; the importer dispatches on a prefix in the regex. One class, two flavors.
> 2. **A generic `BlueprintScopedExporter<T : BlueprintId>` + registry.** The exporter core logic is identical; only the path template and the ID lookup vary. Inject a `PathTemplate` and a `ScopedDefinitionLookup<T>` and you get one implementation per artifact type instead of two.
> 3. **Treat "owner" as a first-class concept end-to-end.** Since the repository already keys on `(BlueprintType, key, versionTag)`, push `BlueprintType` all the way up. Importers then key off `type()` returning something like `FORM_DEFINITION` universally, with the dependency graph expressed via scope rather than via a separate `BUILDING_BLOCK_*` import-type enum. The `ValtimoImportTypes` doubling (CASE_DEFINITION / FORM vs BUILDING_BLOCK_DEFINITION / BUILDING_BLOCK_FORM_DEFINITION) is the symptom that the scope wasn't modeled as data.
> 4. **Strategy for path layout only.** If the divergence is truly just "where on disk does the config bundle live," inject a `BundleLayout` strategy and keep a single importer/exporter. This is the minimum-viable refactor.
> 
> **Why it probably happened**
> 
> My guess: building-block was added after case-scoped deployment was already working, under time pressure, and copy-paste-modify is the lowest-risk way to ship without destabilizing the existing case path. That's a legitimate _tactical_ choice — but it leaves a structural debt, because now every future change to import/export mechanics has to be made in two places, and the `ValtimoImportTypes` enum keeps accreting parallel tokens. Option 1 or 4 above would be a reasonable consolidation pass once the feature is stable.
> 
> Short version: there isn't a strong design reason for the duplication, just a historical one. The persistence layer already proves the right abstraction exists; the upper layers just didn't follow through.

Good point, I created an issue for it to be picked up later: https://github.com/orgs/generiekzaakafhandelcomponent/projects/5/views/1?pane=issue&itemId=176611283&issue=generiekzaakafhandelcomponent%7Cgzac-issues%7C683

Klaas-Ritense (approved):

Klaas-Ritense:
Test results:
- All kinds of random errors when using the ' Energietoeslag Aanvraag' case. But they dont have anything to do with the DMN

PR #661, opened 2026-05-19
GH-446: Dependency upgrades

marijnritense:
Forced Netty to a specific version. Spring hasn't updated yet, so this has to do for now.

Fixes https://github.com/generiekzaakafhandelcomponent/atlas-internal/issues/446

### DIFF
--- a/backend/build.gradle
+++ b/backend/build.gradle
@@ -151,6 +151,7 @@ allprojects {
                 mavenBom "software.amazon.awssdk:bom:${awssdkVersion}"
                 mavenBom "com.fasterxml.jackson:jackson-bom:${jacksonVersion}"
                 mavenBom "org.junit:junit-bom:${junitVersion}"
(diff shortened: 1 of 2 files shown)


backend/build.gradle
```diff
@@ -151,6 +151,7 @@ allprojects {
                 mavenBom "software.amazon.awssdk:bom:${awssdkVersion}"
                 mavenBom "com.fasterxml.jackson:jackson-bom:${jacksonVersion}"
                 mavenBom "org.junit:junit-bom:${junitVersion}"
+                mavenBom "io.netty:netty-bom:4.1.133.Final"
```
Klaas-Ritense, line 155: Can the version number be moved to the gradle.properties file?

marijnritense: I would rather we remove the bom dependency altogether the second spring brings out a new version. It's the only reason why this is there right now.

ivo-ritense (approved):

PR #662, opened 2026-05-19
GH-447: Dependency upgrades

marijnritense:
Forced Netty to a specific version. Spring hasn't updated yet, so this has to do for now.

Fixes https://github.com/generiekzaakafhandelcomponent/atlas-internal/issues/447

### DIFF
--- a/backend/build.gradle
+++ b/backend/build.gradle
@@ -102,6 +102,7 @@ allprojects {
                 mavenBom "software.amazon.awssdk:bom:${awssdkVersion}"
                 mavenBom "com.fasterxml.jackson:jackson-bom:${jacksonVersion}"
                 mavenBom "org.junit:junit-bom:${junitVersion}"
+                // Temporary override, remove on next upgrade of spring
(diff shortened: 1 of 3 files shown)


backend/build.gradle
```diff
@@ -102,6 +102,7 @@ allprojects {
                 mavenBom "software.amazon.awssdk:bom:${awssdkVersion}"
                 mavenBom "com.fasterxml.jackson:jackson-bom:${jacksonVersion}"
                 mavenBom "org.junit:junit-bom:${junitVersion}"
+                mavenBom "io.netty:netty-bom:4.1.133.Final"
```
Klaas-Ritense, line 106: Can the version number be moved to the gradle.properties file?

ivo-ritense (approved):

backend/build.gradle
marijnritense: Same here, we want to get rid of this anyway the minute spring updates its own dependencies and brings out a new version. I don't think there is a point in moving it in this case.

PR #686, opened 2026-05-28
fix: improve error message for case def import

mbritense:
Closes https://github.com/orgs/generiekzaakafhandelcomponent/projects/5/views/3?sliceBy%5Bvalue%5D=WEEK+22&pane=issue&itemId=187113038&issue=generiekzaakafhandelcomponent%7Cgzac-issues%7C711

### DIFF
--- a/backend/case/src/main/kotlin/com/ritense/case/web/rest/CaseDefinitionResource.kt
+++ b/backend/case/src/main/kotlin/com/ritense/case/web/rest/CaseDefinitionResource.kt
@@ -404,7 +404,7 @@ class CaseDefinitionResource(
         @RequestParam("key", required = false) key: String?,
         @RequestParam("name", required = false) name: String?,
         @RequestPart("pluginConfigurationMappings", required = false) pluginConfigurationMappingsJson: String?,
-    ): ResponseEntity<CaseDefinitionImportResponse> {
+    ): ResponseEntity<*> {
         return try {
             val pluginConfigurationMappings: Map<UUID, UUID?>? = pluginConfigurationMappingsJson?.let {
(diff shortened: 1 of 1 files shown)


backend/case/src/main/kotlin/com/ritense/case/web/rest/CaseDefinitionResource.kt
```diff
@@ -419,9 +419,10 @@ class CaseDefinitionResource(
             )
             service.setLatestToActiveIfNoneIsActive()
             ResponseEntity.ok(CaseDefinitionImportResponse(caseDefinitionId))
-        } catch (exception: ImportServiceException) {
+        } catch (exception: Exception) {
```
Klaas-Ritense, line 422: Rather than catching exceptions inline and constructing the error response here, consider leveraging `@ExceptionHandler`. This keeps the controller focused on the happy path and ensures a consistent error response contract across the application.

We already have an `@ExceptionHandler` for `BadRequestAlertException` that returns a standardized error body.  You can reuse it by having `ImportServiceException extends BadRequestAlertException` — the exception will then be translated into a `400 Bad Request` with the shared error format automatically, no `try/catch` needed in the resource.

One important concern with the current approach: `catch (exception: Exception)` is too broad. It catches everything that would otherwise surface as a `500 Internal Server Error` — `NullPointerException`s, database errors, IO failures, etc. — and echoes `exception.message` straight back to the frontend. Those messages often contain sensitive internal details that should never leak to the client. A centralized `@ExceptionHandler` setup avoids this by mapping only known, safe exceptions to user-facing messages and letting everything else fall through to a generic 500 response.

The inline `mapOf("errors" to errorMessage)` also introduces an ad-hoc response shape that diverges from the rest of the API's error format.

PR #688, opened 2026-06-02
Updated dependencies for postgres, netty and tomcat

marijnritense:
Solves https://github.com/generiekzaakafhandelcomponent/atlas-internal/issues/450 and https://github.com/generiekzaakafhandelcomponent/atlas-internal/issues/448


gradle.properties
```diff
@@ -74,13 +74,15 @@ mandrillLutungClientVersion=0.0.8
 apacheHttpclientVersion=4.5.14
 plexusUtilsVersion=3.6.1
 log4jVersion=2.25.4
+nettyVersion=4.1.134.Final
```
ivo-ritense, line 77: Assuming these are temporary overrides, doesn't it make sense to put them in their own block in this file?

ivo-ritense (approved):

PR #717, opened 2026-06-11
Building blocks support nested document properties

Klaas-Ritense:
https://github.com/generiekzaakafhandelcomponent/atlas-internal/issues/478


mbritense:
I suspect there's breaking changes, please verify: 

### What is breaking / behavior-changing

Editor prefill regression for pre-existing process links (the one we hit). BuildingBlockFieldService now emits field names as JSON-pointer paths (/applicantName) instead of flat names (applicantName). Any process link created on next-minor stores flat targets, so when opened in the management UI the target/source dropdowns won't prefill, and because required-field re-sync matches on the new path names, re-saving can create duplicate/empty rows and force re-mapping. There is no data migration (confirmed — no changelog in the diff), so every existing customer building-block link is affected in the UI until re-saved.
Input-source resolution context changed in BuildingBlockCallActivityListener. The old code split pv: vs doc:sources and resolved doc: against an explicitly-derived sourceDocumentId (parent case / parent BB document); it threw IllegalStateException for independent processes with doc: sources. The new code resolves all sources against execution.processInstanceId + execution. Equivalent for the standard parent-case→BB case, but a genuine semantic change for nested-BB and independent-process edge cases (the latter no longer throws).
Dotted targets are now nested paths. toJsonPointer converts .→/, so a target like applicantAddress.city now writes nested {applicantAddress:{city}} instead of a literal flat key "applicantAddress.city". Unlikely to be used, but it's a semantic change.

Klaas-Ritense:
@mbritense Not a breaking change — we still support both /applicantName and applicantName (and the other equivalent notations). The empty-dropdown issue was a pre-existing bug that surfaced when an autodeployment file used one of the alternative-but-valid input mapping notations; the UI just didn't recognize it as matching the field. This is now also fixed in this PR.

mbritense (approved):

PR #772, opened 2026-07-02
787 - Enforce export permission for case list CSV export

Klaas-Ritense:
https://github.com/generiekzaakafhandelcomponent/gzac-issues/issues/787


backend/case/src/main/kotlin/com/ritense/case/service/CaseExporter.kt
```diff
@@ -84,6 +91,17 @@ class CaseExporter(
         return toCsvResponse(caseDefinitionKey, exportableCases, exportableColumns)
     }
 
+    private fun requireExportPermission(caseDefinitionKey: String) {
+        authorizationService.requirePermission(
+            RelatedEntityAuthorizationRequest(
```
marijnritense, line 96: Does this work? This should be an addition to a query, right? Rather than a simple yes/no check.

Klaas-Ritense: Yes, it works — and the query-level filtering you're describing is already in place. The row filtering happens in `searchForExport`, which runs `search(..., EXPORT)` and injects the `export` authorization spec into the query `WHERE` clause (same path `VIEW_LIST` uses for the case list). So the CSV only ever contains rows the user is allowed to export, field conditions included.

`requireExportPermission` isn't the row-level enforcement — it's an intentional coarse gate so a user with no export permission on this case definition gets a 403 instead of a silent empty CSV. For a related-entity check without a concrete document, field conditions are skipped by design (`AuthorizationSpecification.isAuthorizedForRelatedEntity`), so it won't wrongly block conditional-export users.

marijnritense: According to the ticket there's an issue with the button showing up as clickable. Are you going to add that to your pull request, or is that out of scope?

Klaas-Ritense: The button is actually already handled on the frontend — this turned out to be a permission configuration issue, not a missing check.

See: https://ritensebv.slack.com/archives/C1ZHMQGF7/p1782914223122349

The download button is only shown when the user has an applicable `export` permission for that case. The problem was that the permission was written with a document-level condition, which doesn't get evaluated when the frontend checks whether to show the button — so it stayed visible. Rewriting it as a `container` condition scoped to the case definition fixes that, and Nadja confirmed it works on her side.

That's exactly the pattern I've documented in this PR (`container-conditions.md`, "Scoping case list exports"). Combined with the new backend check that returns a 403 for unauthorized requests, the button issue is covered — so I'd keep it in scope here without extra frontend changes. Happy to raise a follow-up ticket if we later decide the config guidance isn't enough.

marijnritense (approved):

documentation/release-notes/13.x.x/13.38.0/README.md
```diff
@@ -55,13 +55,19 @@
   with an empty value and the stored file references were never restored, causing the uploaded file
   to disappear from the UI.
 
+* **Case list export now enforces the export permission**
```
marijnritense: Should now be for 13.39.0

Klaas-Ritense: Changed

PR #776, opened 2026-07-03
gh 755 document actions building block authorization

Klaas-Ritense:
https://github.com/generiekzaakafhandelcomponent/gzac-issues/issues/755

> ZGW document actions such as **view** and **modify** could be incorrectly disabled — even when access control permitted them — for documents uploaded or managed from a building block process. The process business key of a building block process holds the building-block document id rather than the owning case document id, so authorization on `ZgwDocument` (with a container condition on `JsonSchemaDocument`) was evaluated against the wrong document. The owning case document id is now resolved consistently across the ZGW plugins, so document actions are enabled according  to the actual authorization configuration.


backend/zgw/zaken-api/src/main/kotlin/com/ritense/zakenapi/ZakenApiPlugin.kt
```diff
@@ -494,7 +500,7 @@ class ZakenApiPlugin(
         @PluginActionProperty processObjectObjectType: String? = null,
         @PluginActionProperty processObjectRegistration: String? = null,
     ) {
-        val documentId = UUID.fromString(execution.businessKey)
+        val documentId = execution.getJsonSchemaDocumentId()
```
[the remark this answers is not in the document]
Klaas-Ritense: zaakUrlProvider.getZaakUrl(...) accepts both documentId and caseDocumentId. Because the implementation of `getZaakUrl` uses the case-document resolver

backend/zgw/documenten-api/src/main/kotlin/com/ritense/documentenapi/DocumentenApiPlugin.kt
```diff
@@ -219,8 +220,7 @@ class DocumentenApiPlugin(
         @PluginActionProperty processVariableName: String? = null
     ): String {
         val documentUrl = resolveDownloadDocumentUrl(execution)
-        val caseDocumentId = UUID.fromString(execution.businessKey)
-            ?: throw IllegalStateException("Failed to store document. Business key is null.")
+        val caseDocumentId = execution.getJsonSchemaDocumentId()
```
ivo-ritense: I think we need to look critically at this and see if this should not use `CaseDocumentResolver` in this and similar situations in the document en zaken api plugin

backend/process-document/src/main/kotlin/com/ritense/processdocument/helper/GetJsonSchemaDocumentHelper.kt
```diff
@@ -0,0 +1,209 @@
...
+
+    private fun joinToString(vararg entries: Pair<String, String?>): String {
+        val details = entries
+            .filter { !it.second.isNullOrBlank() }
+            .joinToString(", ") { "${it.first}: '${it.second}'" }
+        return if (details.isEmpty()) "" else " $details"
+    }
+
+    private fun String?.toUUIDOrNull(): UUID? {
+        return if (this != null && UUID_REGEX.toRegex().matchEntire(this) != null) {
```
ivo-ritense: This compiles the regex every time it's used. Can we not store a compiled version somewhere?

ivo-ritense (approved):

backend/zgw/documenten-api/src/main/kotlin/com/ritense/documentenapi/DocumentenApiPlugin.kt
ivo-ritense: I think this important to look at, but it's not the scope of this PR. Looking at it more closely, it's mostly used to get the zaak url for the document, which in the `ZaakUrlProvider` already uses `CaseDocumentResolver`.

backend/process-document/src/main/kotlin/com/ritense/processdocument/helper/GetJsonSchemaDocumentHelper.kt
```diff
@@ -0,0 +1,209 @@
...
+import org.operaton.bpm.engine.delegate.DelegateTask
+import org.operaton.bpm.engine.delegate.VariableScope
+import org.operaton.bpm.engine.externaltask.ExternalTask
+import org.operaton.bpm.engine.externaltask.LockedExternalTask
+import org.operaton.bpm.engine.history.HistoricCaseInstance
+import org.operaton.bpm.engine.history.HistoricProcessInstance
+import org.operaton.bpm.engine.runtime.CaseInstance
+import org.operaton.bpm.engine.runtime.ProcessInstance
+
+object GetJsonSchemaDocumentHelper {
```
ivo-ritense, line 33: Do we want to make a test for this? I know part of it is indirectly covered by other tests, but it might be worth it to validate these assumptions.

Klaas-Ritense: Ive added a test

backend/process-document/src/main/kotlin/com/ritense/processdocument/helper/GetJsonSchemaDocumentHelper.kt
Klaas-Ritense: Changed

backend/zgw/documenten-api/src/main/kotlin/com/ritense/documentenapi/DocumentenApiPlugin.kt
Klaas-Ritense: Thx, Ive changed it so that it does now resolve the documentId from the case. Resolving the case-document-id here is correct because this caseDocumentId is later used by the `DocumentenApiClient` to check permissions of the `ZgwDocument`. And according to the original Github issue, people expect that the  `ZgwDocument.caseDocumentId` is always the document from the case


In this PR Ive introduced this distinction:
`documentId` is either from builing block or case
`caseDocumentId` always from a case
`buildingBlockDocumentId` always from a building block

PR #786, opened 2026-07-08
Story/493 task form plugin surface

mbritense:
Closes https://github.com/orgs/generiekzaakafhandelcomponent/projects/5/views/3?sliceBy%5Bvalue%5D=WEEK+28&pane=issue&itemId=199437110&issue=generiekzaakafhandelcomponent%7Catlas-internal%7C493


ivo-ritense:
1. I'm worried that the way of executing a task from a plugin is not very user friendly for the developer of a plugin. It looks like there's not much of a framework in place. This means the develop has to know the different steps involved and link them up themselves based on probably documentation. It involves as far as i can see:
- A frontend component for the task form.
- Wiring this up to do a call to the plugin.
- A request handler indistinguishable from other interactions with the plugin.
- In that handler you do an api call to `/api/v1/task/${taskId}/complete`.

I do like the flexibility it provides for all kinds of scenario's though.

2. I'm not sure how this should deal with information. Currently most task forms contains some data input. When this is directly used there's no issue, but sending data back to be used in the valtimo case is tough. The plugin developer doesn't know what the data model of the case definition is that it will be used in (or they do and the plugin is not as re-usable). I would expect some kind of optional mapping maybe where you can bind the resulting data from the form to some data in the case?

ivo-ritense (approved):

PR #810, opened 2026-07-14
Opensearch for document queries + generic case search

marijnritense:


frontend/projects/valtimo/document/src/lib/services/document.service.ts
```diff
@@ -1,19 +1,21 @@
...
+ *  * you may not use this file except in compliance with the License.
+ *  * You may obtain a copy of the License at
+ *  *
+ *  * https://joinup.ec.europa.eu/collection/eupl/eupl-text-eupl-12
+ *  *
+ *  * Unless required by applicable law or agreed to in writing, software
+ *  * distributed under the License is distributed on an "AS IS" basis,
+ *  * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  * See the License for the specific language governing permissions and
+ *  * limitations under the License.
```
sofiaIvarsRitense: Could you check the headers modified in this pull request, please? It looks like they are incorrectly re-nested

frontend/projects/valtimo/components/src/lib/components/carbon-list/carbon-list.component.ts
```diff
@@ -308,21 +327,9 @@ export class CarbonListComponent implements OnInit, AfterViewInit, OnDestroy {
 
     this._subscriptions.add(
       this.searchFormControl.valueChanges
-        .pipe(debounceTime(500))
+        .pipe(debounceTime(2000))
```
sofiaIvarsRitense: This _debounceTime(500) → debounceTime(2000)_ is used by a lot of templates, so it affects the search box in all of those lists. If the global search needs different timing, it could be an @Input() searchDebounceTime = 500 rather than changing the global value

frontend/projects/valtimo/components/src/lib/components/search-input-with-validation/search-input-with-validation.component.ts
```diff
@@ -0,0 +1,147 @@
...
+import {CommonModule} from '@angular/common';
+import {FormsModule} from '@angular/forms';
+import {IconModule, IconService} from 'carbon-components-angular';
+import {Search16, Close16} from '@carbon/icons';
+
+interface TextSegment {
+  text: string;
+  isFieldName: boolean;
+  isInvalid: boolean;
+}
```
sofiaIvarsRitense: This interface should be in a different file

frontend/projects/valtimo/components/src/lib/components/search-input-with-validation/search-input-with-validation.component.ts
```diff
@@ -0,0 +1,147 @@
...
+
+@Component({
+  standalone: true,
+  selector: 'valtimo-search-input-with-validation',
+  templateUrl: './search-input-with-validation.component.html',
+  styleUrls: ['./search-input-with-validation.component.scss'],
+  changeDetection: ChangeDetectionStrategy.OnPush,
+  imports: [CommonModule, FormsModule, IconModule],
+})
+export class SearchInputWithValidationComponent {
```
sofiaIvarsRitense: I think this component is not used anywhere

frontend/projects/valtimo/components/src/lib/components/carbon-list/carbon-list.component.ts
```diff
@@ -759,4 +777,288 @@ export class CarbonListComponent implements OnInit, AfterViewInit, OnDestroy {
...
+    }
+
+    return segments;
+  }
+
+  public showAutocomplete = false;
+  public filteredSuggestions: SearchField[] = [];
+  public selectedSuggestionIndex = -1;
+  public autocompleteLeft = 0;
+  private _searchInputElement: HTMLInputElement | null = null;
```
sofiaIvarsRitense: This fields should be at the top

frontend/projects/valtimo/components/src/lib/components/carbon-list/carbon-list.component.ts
```diff
@@ -178,6 +180,18 @@ export class CarbonListComponent implements OnInit, AfterViewInit, OnDestroy {
...
+  @Input() set initialSearchValue(value: string | null) {
+    if (this._initialSearchValue !== value) {
+      this._initialSearchValue = value;
+      this.searchFormControl.setValue(value || '', {emitEvent: false});
+      this._lastExecutedSearch = value;
+      this._searchActive = !!value;
+    }
+  }
+  private _initialSearchValue: string | null = null;
+  public _searchActive = false;
```
sofiaIvarsRitense:
```suggestion
  public searchActive = false;
```

It is missing 'public' and normally '_' is used for private ones

frontend/projects/valtimo/components/src/lib/components/carbon-list/carbon-list.component.scss
```diff
@@ -135,4 +135,95 @@ td:first-child {
...
+  white-space: pre;
+  pointer-events: none;
+  overflow: hidden;
+
+  &__invalid {
+    text-decoration: underline wavy var(--cds-support-error, #da1e28);
+    text-decoration-skip-ink: none;
+    text-underline-offset: 3px;
+  }
+
```
sofiaIvarsRitense: Please remove the extra space

backend/apps/dev/src/main/resources/config/application.yml
```diff
@@ -1,6 +1,9 @@
 logging:
     file:
         name: /tmp/spring.log
+    level:
+        com.ritense.document.opensearch: DEBUG
+        org.opensearch.client.RestClient: WARN
```
ivo-ritense: What happens when this is not set? Do we get a lot of spam about this? What if opensearch is not used, is that still the case?

backend/apps/dev/src/main/resources/logback-spring.xml
```diff
@@ -22,6 +22,7 @@
     <include resource="config/logging/logback/valtimo-database-appender.xml" />
 
     <logger name="com.ritense.valtimo" level="#logback.loglevel#"/>
+    <logger name="com.ritense.document.mongodb" level="DEBUG"/>
```
ivo-ritense: Remove this

backend/apps/dev/build.gradle
```diff
@@ -36,6 +36,7 @@ dependencies {
     implementation(platform(project(":backend:dependencies:valtimo-dependency-versions")))
 
     implementation(project(":backend:dependencies:valtimo-gzac-dependencies"))
+    implementation(project(":backend:case-opensearch"))
```
ivo-ritense: Should this be added to the other apps? I think so. What about the valtimo/gzac dependencies?

backend/case-opensearch/src/main/kotlin/com/ritense/document/opensearch/authorization/mapper/JsonSchemaDocumentCaseDefinitionOpenSearchMapper.kt
```diff
@@ -0,0 +1,73 @@
...
+import com.ritense.document.opensearch.authorization.OpenSearchPermissionConditionTranslator.Companion.applyOperator
+import com.ritense.valtimo.contract.authorization.CurrentUserExpressionHandler
+import org.opensearch.index.query.QueryBuilder
+import org.opensearch.index.query.QueryBuilders
+
+/**
+ * Handles [com.ritense.authorization.permission.condition.ContainerPermissionCondition]
+ * where the container resource type is [CaseDefinition].
+ *
+ * Field paths mirror the MongoDB version: `definitionId.blueprintId.*`.
```
ivo-ritense: Reference to mongo that will be removed/has been removed.

backend/case-opensearch/src/main/kotlin/com/ritense/document/opensearch/authorization/mapper/JsonSchemaDocumentDefinitionOpenSearchMapper.kt
```diff
@@ -0,0 +1,69 @@
...
+import com.ritense.document.opensearch.authorization.OpenSearchPermissionConditionTranslator.Companion.andAll
+import com.ritense.document.opensearch.authorization.OpenSearchPermissionConditionTranslator.Companion.applyOperator
+import com.ritense.valtimo.contract.authorization.CurrentUserExpressionHandler
+import org.opensearch.index.query.QueryBuilder
+
+/**
+ * Handles [com.ritense.authorization.permission.condition.ContainerPermissionCondition]
+ * where the container resource type is [JsonSchemaDocumentDefinition].
+ *
+ * Field paths mirror the MongoDB version: `definitionId.name` and `definitionId.version`.
```
ivo-ritense: Again mongo reference.

backend/case-opensearch/src/main/kotlin/com/ritense/document/opensearch/authorization/OpenSearchAuthorizationEntityMapper.kt
```diff
@@ -0,0 +1,41 @@
...
+/**
+ * OpenSearch equivalent of [com.ritense.authorization.AuthorizationEntityMapper].
+ *
+ * Translates a [com.ritense.authorization.permission.condition.ContainerPermissionCondition]
+ * on entity type [TO] into an OpenSearch [QueryBuilder] that filters [FROM] documents.
+ *
+ * Implement this interface and register the implementation as a Spring bean to add support
+ * for a new container relationship without modifying the core translator.
+ */
+interface OpenSearchAuthorizationEntityMapper<FROM : Any, TO : Any> {
```
ivo-ritense, line 31: Does this mean every link to another entity has to have this mapper as well? Doesn't seem to scale well with new entities.

sofiaIvarsRitense:
- If I open the “Start reindex” modal, fill in the fields, and close it without saving, then reopen the modal, the “Modified after” field is still populated with the previously filled date. The form should be reset when the modal is closed without saving
- When the “Start reindex” modal is opened, the “Start reindex” button is already enabled. It should remain disabled until all required fields have been completed
- Date fields such as “Modified after” and “Started/Finished” use a different date format from the rest of the application
- It is possible to submit the form with the default “All definitions” option selected. However, after expanding the accordion, the “Document definition” field is displayed as empty '-'

backend/apps/dev/build.gradle
marijnritense: I'll add it to both of them

backend/case-opensearch/src/main/kotlin/com/ritense/document/opensearch/authorization/OpenSearchAuthorizationEntityMapper.kt
marijnritense: I tried to explore some alternatives:
- Reusing JPA metadata (can't be done without AST parsing, so that's going to make things a lot more complicated)
- Annotations on entities (can be done, but would still end up requiring a mapper. Kind of defeats the purpose)
- 1 generic mapper and then special registries. These registries would be fairly small, e.g.:

```kotlin
     OpenSearchFieldMappingRegistry().apply {
         register(JsonSchemaDocumentDefinition::class.java, mapOf(
             "id.name" to "definitionId.name",
             "id.version" to "definitionId.version"
         ))
         register(CaseDefinition::class.java, mapOf(
             "id.key" to "definitionId.blueprintId.blueprintKey",
             "id.versionTag" to "definitionId.blueprintId.blueprintVersionTag"
         ))
     }
```
Would make it slightly easier to work with, but I'm not sure it's worth the effort. You still need to register stuff.

Which one do you like most, or do you feel like the current way of doing it is fine?

backend/case-opensearch/src/main/kotlin/com/ritense/document/opensearch/authorization/mapper/JsonSchemaDocumentDefinitionOpenSearchMapper.kt
marijnritense: fixed

backend/case-opensearch/src/main/kotlin/com/ritense/document/opensearch/authorization/mapper/JsonSchemaDocumentCaseDefinitionOpenSearchMapper.kt
marijnritense: fixed

backend/apps/dev/src/main/resources/logback-spring.xml
marijnritense: fixed

frontend/projects/valtimo/components/src/lib/components/carbon-list/carbon-list.component.scss
marijnritense: fixed

frontend/projects/valtimo/components/src/lib/components/carbon-list/carbon-list.component.ts
marijnritense: fixed

frontend/projects/valtimo/components/src/lib/components/search-input-with-validation/search-input-with-validation.component.ts
marijnritense: removed

frontend/projects/valtimo/components/src/lib/components/search-input-with-validation/search-input-with-validation.component.ts
marijnritense: I removed it since it wasn't used outside of the unused component

frontend/projects/valtimo/document/src/lib/services/document.service.ts
marijnritense: Fixed

frontend/projects/valtimo/components/src/lib/components/carbon-list/carbon-list.component.ts
marijnritense: Fixed

frontend/projects/valtimo/components/src/lib/components/carbon-list/carbon-list.component.ts
marijnritense: moved!

marijnritense:
> * If I open the “Start reindex” modal, fill in the fields, and close it without saving, then reopen the modal, the “Modified after” field is still populated with the previously filled date. The form should be reset when the modal is closed without saving
> 
>     * When the “Start reindex” modal is opened, the “Start reindex” button is already enabled. It should remain disabled until all required fields have been completed
> 
>     * Date fields such as “Modified after” and “Started/Finished” use a different date format from the rest of the application
> 
>     * It is possible to submit the form with the default “All definitions” option selected. However, after expanding the accordion, the “Document definition” field is displayed as empty '-'

Should all be done now! We discussed the date fields separately, and the format is actually consistent.

backend/apps/dev/src/main/resources/config/application.yml
marijnritense: The ritense one doesn't matter. I've also changed it so it's configured in `logback-spring.xml` instead.

We do in fact get spam when OpenSearch is used at INFO level (the default), this is why I've lowered it. When not using OpenSearch, there will be no spam.

sofiaIvarsRitense (approved):

backend/case/src/main/java/com/ritense/document/service/impl/JsonSchemaDocumentSearchService.java
```diff
@@ -812,4 +906,75 @@ void apply(
...
+        if (path.startsWith(DOC_PREFIX)) return path.substring(DOC_PREFIX.length());
+        if (path.startsWith(CASE_PREFIX)) return path.substring(CASE_PREFIX.length());
+        return path;
+    }
+
+    private String buildLikePattern(String value, boolean quoted, SearchFieldMatchType matchType) {
+        if (quoted || matchType != SearchFieldMatchType.LIKE) {
+            return value;
+        }
+        return "%" + value + "%";
```
ivo-ritense: What if the searched value contains a %. Should it be escaped?

backend/case/src/main/java/com/ritense/document/service/impl/JsonSchemaDocumentSearchService.java
```diff
@@ -394,6 +397,97 @@ private void buildQueryWhere(
...
+                            } else {
+                                var likePattern = buildLikePattern(term.value(), term.quoted(), field.getMatchType());
+                                qualifiedPredicates.add(cb.like(
+                                    cb.lower(documentRoot.get(fieldPath).as(String.class)),
+                                    likePattern.toLowerCase()
+                                ));
+                            }
+                        }
+                    } else {
+                        var likePattern = "%" + term.value().toLowerCase() + "%";
```
ivo-ritense: Shouldn't this also use the buildLikePattern method?

backend/case-opensearch/src/main/kotlin/com/ritense/document/opensearch/service/DocumentOpenSearchReindexService.kt
```diff
@@ -0,0 +1,337 @@
...
+        }
+        return run.id
+    }
+
+    /**
+     * Runs the chunked, resumable re-index loop for [runId] over the documents matching [scope].
+     * Each DB page is read in its own short read-only transaction (keeping snapshots short) and the
+     * persistence context is cleared after every page. Returns the number of documents processed.
+     */
+    open fun reindex(runId: UUID, scope: ReindexRequest): Long {
```
ivo-ritense: This is a confusing method. If this is used to resume a previously running re-index it shouldn't need a ReindexRequest because it should use the information from the initial run. If not, what is the difference with the start method?

backend/apps/dev/src/main/kotlin/com/ritense/gzac/opensearch/InsecureSslOpenSearchClientConfig.kt
```diff
@@ -0,0 +1,74 @@
...
+
+/**
+ * Dev-only configuration that disables SSL certificate verification for OpenSearch
+ * and configures basic authentication.
+ * Allows connecting to OpenSearch with self-signed certificates in local development.
+ *
+ * DO NOT use this configuration in production.
+ */
+@Configuration
+class InsecureSslOpenSearchClientConfig {
```
ivo-ritense, line 39: Is this required? Feels bad to test with different code in development than we would use in production

backend/apps/dev/build.gradle
```diff
@@ -51,6 +51,9 @@ dependencies {
     implementation "org.springframework.boot:spring-boot-starter-webflux"
     implementation "org.springframework.boot:spring-boot-starter-security"
 
+    // OpenSearch client for InsecureSslOpenSearchClientConfig
+    implementation "org.opensearch.client:spring-data-opensearch-starter:${springDataOpenSearchVersion}"
```
ivo-ritense, line 55: Should this not be a dependency of valtimo? Do implementations need to include this speficically? I thought we tested with all dependencies included but opensearch disabled, so it shouldn't break anything.

backend/apps/gzac/src/main/resources/config/application.yml
```diff
@@ -137,6 +137,11 @@ server:
         mime-types: text/html,text/xml,text/plain,text/css, application/javascript, application/json
         min-response-size: 1024
 
+opensearch:
+    uris: ${OPENSEARCH_URIS:http://localhost:9200}
```
ivo-ritense: Not sure we want to default to a localhost address in a docker container meant for production use.

backend/case-opensearch/src/main/kotlin/com/ritense/document/opensearch/authorization/OpenSearchPermissionConditionTranslator.kt
```diff
@@ -0,0 +1,225 @@
...
+        }
+        return jpaFallback(cond)
+    }
+
+    /**
+     * Fallback for [ContainerPermissionCondition] types that have no registered
+     * [OpenSearchAuthorizationEntityMapper]. Uses JPA to find matching document IDs and
+     * returns an `ids` query.
+     */
+    private fun jpaFallback(cond: ContainerPermissionCondition<*>): QueryBuilder {
```
ivo-ritense, line 124: Not 100% sure, but i couldn't find a test that covers this fall back scenario

backend/case-opensearch/src/main/kotlin/com/ritense/document/opensearch/authorization/OpenSearchPermissionConditionTranslator.kt
```diff
@@ -0,0 +1,225 @@
...
+import com.ritense.authorization.role.Role
+import com.ritense.document.domain.impl.JsonSchemaDocument
+import com.ritense.document.repository.impl.JsonSchemaDocumentRepository
+import com.ritense.valtimo.contract.authorization.CurrentUserExpressionHandler
+import io.github.oshai.kotlinlogging.KotlinLogging
+import org.opensearch.index.query.BoolQueryBuilder
+import org.opensearch.index.query.QueryBuilder
+import org.opensearch.index.query.QueryBuilders
+
+class OpenSearchPermissionConditionTranslator(
```
ivo-ritense, line 39: Can you add some tests for this class?

backend/apps/dev/build.gradle
marijnritense: It's needed because of the InsecureSslOpenSearchClientConfig class.

backend/apps/gzac/src/main/resources/config/application.yml
marijnritense: Agreed, I'll leave them blank

backend/case-opensearch/src/main/kotlin/com/ritense/document/opensearch/authorization/OpenSearchPermissionConditionTranslator.kt
marijnritense: Added

backend/case-opensearch/src/main/kotlin/com/ritense/document/opensearch/authorization/OpenSearchPermissionConditionTranslator.kt
marijnritense: Done

backend/apps/dev/src/main/kotlin/com/ritense/gzac/opensearch/InsecureSslOpenSearchClientConfig.kt
marijnritense: The setup required here is too much in my opinion, especially since we don't need to test opensearch itself, just the things we're adding to/around it. The only real alternative is disabling the security plugin for local dev, but then that moves us even further away from the actual scenario, as we'd be using plain HTTP.

backend/case-opensearch/src/main/kotlin/com/ritense/document/opensearch/service/DocumentOpenSearchReindexService.kt
marijnritense: Changed it to hopefully make more sense.

backend/case/src/main/java/com/ritense/document/service/impl/JsonSchemaDocumentSearchService.java
marijnritense: Fixed

backend/case/src/main/java/com/ritense/document/service/impl/JsonSchemaDocumentSearchService.java
marijnritense: Fixed

ivo-ritense (approved):

PR #828, opened 2026-07-21
GH-817: Valueresolver does not resolve the case properties in a building block

marijnritense:
Changed `case:` value resolver to always retrieve data from the case, even within the context of a building block. When within the context of a building block, no data can be written to a case.

Solves https://github.com/generiekzaakafhandelcomponent/gzac-issues/issues/817


backend/process-document/src/main/kotlin/com/ritense/processdocument/resolver/DocumentTableValueResolver.kt
```diff
@@ -65,9 +65,13 @@ class DocumentTableValueResolver(
     }
 
     override fun handleValues(processInstanceId: String, variableScope: VariableScope?, values: Map<String, Any?>) {
-        val documentId =
-            processDocumentService.getDocumentId(OperatonProcessInstanceId(processInstanceId), variableScope)
-        handleValues(documentId, values)
+        val operatonProcessInstanceId = OperatonProcessInstanceId(processInstanceId)
+        val documentId = processDocumentService.getDocumentId(operatonProcessInstanceId, variableScope)
```
ivo-ritense, line 69: Klaas just introduced a help for this kind of thing https://github.com/valtimo-platform/valtimo/blob/next-minor/backend/process-document/src/main/kotlin/com/ritense/processdocument/helper/GetJsonSchemaDocumentHelper.kt#L52

ivo-ritense (approved):

PR #832, opened 2026-07-21
662 - Fix ESC key not reliably closing Carbon modals

maarten-ritense:
### Describe the changes
Link to the related Github issue: generiekzaakafhandelcomponent/gzac-issues#662

**1. ESC key did not reliably close Carbon modals.**
ESC-to-close was handled solely by Carbon's built-in bubbling `keydown` listener on `cds-modal`, so it only fired while keyboard focus was *inside* the modal. As soon as focus landed on `<body>` — which happens when the user clicks a non-focusable part of the modal, and in development builds happens immediately on open because `ValtimoCdsModalDirective.preventModalCloseButtonTooltip()` blurs the auto-focused close button — ESC did nothing. Users experienced this as a task/modal that could not be closed (or reopened) without a full page refresh.

`ValtimoCdsModalDirective` (shared by the Carbon modals) now handles ESC itself via a **focus-independent, document-level capture-phase** `keydown` listener. For the top-most visible modal it preempts Carbon's handler (`stopImmediatePropagation`) and simulates a click on that modal's own close (X) button, so ESC runs through the exact same `(closeSelect)` → close path as the X button. This also avoids Carbon's ESC side effects (mutating `open` directly outside the `[open]` binding and calling `modalService.destroy()`, which can affect an unrelated imperatively-created modal).

**2. Several modals never wired up the shared `valtimoCdsModal` behaviour.**
While fixing (1), we found many modals that write `<cds-modal ... valtimoCdsModal>` in their template but never made `ValtimoCdsModalDirective` available — standalone components omitted it from `imports`, and one NgModule omitted it from its `imports`. On those modals the attribute was inert, so they missed the body scroll lock, `min-content-height` sizing, close-button tooltip suppression, and (after change 1) the ESC-to-close handling. The most visible symptom was the ZGW "Connected zaak type" modal, which showed a lingering "Close modal" tooltip. All affected components/modules now import the directive: the ZGW connected-zaak-type and Documenten API column/upload-field/tag modals, `failed-notification-detail`, the process/form/building-block upload modals, the widget-management divider + wizard modals, the decision deploy modal, and the access-control role modals.

**3. ESC while a dropdown/combo-box is open now closes only the menu.**
When a Carbon dropdown or combo-box (`cds-dropdown` / `cds-combo-box`, including the `v-select` and `valtimo-value-path-selector` wrappers) is open inside a modal, ESC should close only that menu, not the modal. Carbon closes the menu itself on ESC, but `cds-combo-box` — unlike `cds-dropdown` — does not stop the event, so it would bubble up to the modal's ESC handler and close it too. So for an open list-box the directive does **not** preempt/close the modal (change 1 only applies when no menu is open); it lets Carbon close the menu — its own handler runs on the real event in Angular's zone, keeping state and DOM in sync — and only **shields the modal** by adding a one-time `keydown` listener on the list-box host that stops propagation after Carbon's handler. A second ESC (menu closed) then closes the modal. Detection is document-wide because with `appendInline=false` Carbon detaches the expanded menu — and moves focus into it — outside the modal, so neither the marker nor the event target is inside the modal.

Specify the code branch location: `bugfix/662-esc-key-modal-close` (base: `next-minor`)

**Relevant comments:**
> The ESC fix lives in the shared directive, so it applies to every correctly-wired Carbon modal with no template changes. Modal ownership is resolved via `topModal.closest('cds-modal') === host` (not `contains`), which is required because the task-detail modal embeds nested `cds-modal`s from its form content — nested/stacked modals therefore close top-first and the correct (own) close button is used. When a task opens in the right-hand panel instead of a modal there is no visible `cds-modal`, so ESC is a natural no-op. The release notes for this change target **13.39.0** (it is not part of the 13.38.0 release).

### Breaking changes

### Documentation

- [ ] Yes
- [x] Not applicable

### Tests
- [x] Yes
- [ ] Not applicable

- [ ] Yes
- [x] Not applicable

Describe the testing steps
- [x] Open a user task modal (from a case detail page and from the task list) and a config modal (e.g. ZGW "Connected zaak type"); press ESC without interacting first — the modal closes
- [x] Open a modal, click a non-focusable area (focus moves to `<body>`), press ESC — the modal still closes
- [x] After closing with ESC, immediately reopen — no stuck overlay, page stays interactive (no refresh needed)
- [x] Confirm the ZGW "Connected zaak type" modal no longer shows a lingering "Close modal" tooltip and behaves like the other modals
- [x] Confirm the close (X) button, backdrop and nested/confirmation modals still behave as before (nested modals close top-first)
- [x] Open a modal containing a dropdown/combo-box (e.g. the task-list column modal "Path" and "Display type" fields, or the ZGW "Object management configuration" select); open the menu and press ESC — only the menu closes and the modal stays open; press ESC again — the modal closes

### Security
- [ ] Yes
- [x] Not applicable

- [ ] Yes
- [x] Not applicable

- [ ] Yes
- [x] Not applicable

### Dependencies
- [ ] Yes
- [x] Not applicable


frontend/projects/valtimo/components/src/lib/directives/valtimo-cds-modal/valtimo-cds-modal.directive.spec.ts
```diff
@@ -0,0 +1,126 @@
...
+import {Component, CUSTOM_ELEMENTS_SCHEMA} from '@angular/core';
+import {ComponentFixture, TestBed} from '@angular/core/testing';
+import {ValtimoCdsModalDirective} from './valtimo-cds-modal.directive';
+
+@Component({
+  standalone: true,
+  imports: [ValtimoCdsModalDirective],
+  schemas: [CUSTOM_ELEMENTS_SCHEMA],
+  template: `
+    <cds-modal valtimoCdsModal>
```
sofiaIvarsRitense: Please move this part to a html file

frontend/projects/valtimo/components/src/lib/directives/valtimo-cds-modal/valtimo-cds-modal.directive.spec.ts
```diff
@@ -0,0 +1,126 @@
+/*
+ * Copyright 2015-2025 Ritense BV, the Netherlands.
```
sofiaIvarsRitense: 2026

frontend/projects/valtimo/components/src/lib/directives/valtimo-cds-modal/valtimo-cds-modal.directive.spec.ts
```diff
@@ -0,0 +1,154 @@
...
+
+    document.dispatchEvent(event);
+
+    // Modal is protected (event stopped, close button not clicked) and the open menu is closed via
+    // a synthetic outside click on the body — which is how every Carbon list-box closes.
+    expect(clickSpy).not.toHaveBeenCalled();
+    expect(stopSpy).toHaveBeenCalled();
+    expect(bodyClickSpy).toHaveBeenCalled();
+    document.body.removeEventListener('click', bodyClickSpy);
+  });
```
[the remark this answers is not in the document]
maarten-ritense: This is a false positive — the "return without consuming ESC and let Carbon close its own menu" approach was tried first and verified not to work for `cds-combo-box`.

Carbon's two list-box components handle ESC inconsistently:

- `cds-dropdown` (`dropdown.component`): on ESC while open it calls `event.stopImmediatePropagation()` (comment in source: *"don't unintentionally close modal if inside of it"*), so the modal is spared and bailing works.
- `cds-combo-box` (`combobox.component`, which `v-select` wraps — the reported "Object management configuration" field): its host `hostkeys()` handler calls `closeDropdown()` on ESC but does **not** stop propagation. So if we return without consuming, the ESC bubbles on to Carbon's modal `handleKeyboardEvent`, which runs `stopImmediatePropagation()` + `this.open = false` + `modalService.destroy()` and closes the modal too.

On top of that, both components render with `appendInline=false` here, so the expanded `.cds--list-box--expanded` element is detached to `document.body`, outside the modal.

I confirmed the bail approach live: with it, pressing ESC once on the open combo-box closed **both** the menu and the modal (exactly the regression Sofia reported). The current approach — `stopImmediatePropagation()` to protect the modal, then a synthetic outside click to close the open menu (which every Carbon list-box listens for via its document capture `outsideClick` handler) — was verified end-to-end in the running app for both the `cds-combo-box` (Object management configuration) and the `cds-dropdown` (Display Type in the column modal): first ESC closes only the menu, second ESC closes the modal. The tests assert that actual, working behavior, so I'm keeping them as-is.

sofiaIvarsRitense (approved):

sofiaIvarsRitense (approved):

PR #839, opened 2026-07-22
338 - log and reliably (re)register Notificatie API abonnementen on plugin config changes

maarten-ritense:
### Describe the changes
Link to the related Github issue: generiekzaakafhandelcomponent/gzac-issues#338

**Notificatie API subscriptions were created/updated/deleted silently, and not always reliably.**
When a plugin that relies on the Notificatie API (e.g. the Verzoek plugin) is added, modified, or
removed, Valtimo (re)registers its subscription ("abonnement") against the Notificatie API. This
happened without any log entry — a successful `POST`/`DELETE` to the abonnementen endpoint left no
trace, and failures were swallowed by the retry loop until the final attempt threw, so an operator
who changed a plugin and saw no subscription appear had nothing to go on. On top of that, the
(re)registration was triggered from `PluginService` *mid-transaction* (via the empty
`PluginsDeployedEvent`) and handled synchronously, so the external HTTP calls could run before the
plugin configuration change was committed — leaving the remote subscription inconsistent with the
stored configuration if the transaction rolled back.

**What changed:**
1. **Dedicated events.** Added `PluginConfigurationCreatedEvent` and
   `PluginConfigurationUpdatedEvent` (in `com.ritense.plugin.events`), each carrying the affected
   `PluginConfiguration`. `PluginService` publishes them from `createPluginConfiguration` and
   `updatePluginConfiguration`. They mirror the existing `PluginConfigurationDeletedEvent`.
2. **Reliable, scoped trigger.** `PluginsDeployedEventListener` now (re)registers abonnementen from
   typed create/update/delete listeners using `@TransactionalEventListener(phase = AFTER_COMMIT)` —
   matching how the same class already handles process-link events. The external Notificatie API
   calls therefore only run once the plugin configuration change is committed. Because these
   lifecycle events fire for *every* plugin, each handler is now gated to configurations that are a
   `NotificatiesApiListener` (checked via the plugin definition class name), so unrelated plugin
   changes no longer trigger remote reconciliation. Startup registration is unchanged (still driven
   by `ApplicationFullyReadyEvent`).
3. **Deletion reconciliation.** When a Notificaties API plugin configuration is deleted, its remote
   abonnement and local `NotificatiesApiAbonnementLink` are now explicitly removed — previously they
   were orphaned when the deleted configuration was the last one for its Notificatie API URL — after
   which the remaining configurations are re-registered.
4. **Logging.** Successful create/update/delete of a subscription now logs at `INFO`; retry failures
   log at `WARN` per attempt and `ERROR` on the final failure, including the reason; and skipping
   because registration is disabled (`valtimo.zgw.register-abonnementen=false`) is now logged
   instead of being silent. Every entry is scoped with the plugin configuration id via the logging
   context (`withLoggingContext`), and — for changes triggered from an HTTP request — the existing
   request `valtimoCorrelationId` is already present in the MDC, so a subscription operation can be
   traced back to the call that caused it.

Specify the code branch location: `story/338-notification-api-logging` (base: `next-minor`)

**Relevant comments:**
> The empty `PluginsDeployedEvent` is **no longer published by `PluginService`** on plugin
> configuration create/update — dropped after confirmation it is safe to remove. Its only in-repo
> consumer (`PluginsDeployedEventListener`) now reacts to the dedicated
> `PluginConfigurationCreated`/`Updated`/`DeletedEvent` events instead. The event class itself
> remains in the public contract module (`com.ritense.valtimo.contract.event`) and is still published
> by plugin auto-deployment, so any downstream implementation subscribing to it keeps receiving it
> there; such consumers should migrate to the dedicated events. The listener class keeps its (now
> slightly historical) name to avoid churn in the auto-configuration and in other modules'
> integration-test base classes that reference it. Only non-sensitive identifiers are logged
> (abonnement id from the URL, plugin configuration id) — the abonnement `auth` secret is never
> logged.

### Breaking changes

### Documentation

- [x] Yes
- [ ] Not applicable

### Tests
- [x] Yes
- [ ] Not applicable

- [ ] Yes
- [x] Not applicable

Describe the testing steps
- [x] `./gradlew :backend:plugin:test --tests "com.ritense.plugin.service.PluginServiceTest"` — create/update publish the new events (green)
- [x] `./gradlew :backend:zgw:notificaties-api:test --tests "com.ritense.notificatiesapi.PluginsDeployedEventListenerTest"` — success/failure/disabled log entries and correlation id asserted, plus deletion cleanup, the untracked/disabled deletion cases, and that unrelated plugin changes are ignored (green)
- [x] Run the app, add / modify / delete a Verzoek (or Portaaltaak) plugin configuration; confirm the console shows `Successfully created/updated/deleted abonnement …` with the plugin configuration id in the MDC, and that the subscription is (re)registered only after the change is committed
- [x] Delete a Notificaties API plugin configuration and confirm its remote abonnement and local link are removed; delete a Verzoek listener and confirm the abonnement is updated (kanalen) rather than deleted
- [x] Confirm no subscription secret (`auth`) appears in the logs

### Security
- [x] Yes
- [ ] Not applicable

- [ ] Yes
- [x] Not applicable

- [ ] Yes
- [x] Not applicable

### Dependencies
- [ ] Yes
- [x] Not applicable


documentation/release-notes/13.x.x/13.39.0/migration.md
Klaas-Ritense, line 1: This file is not needed

documentation/release-notes/13.x.x/13.39.0/README.md
```diff
@@ -6,16 +6,30 @@
 
 ## New Features
 
-* **New feature title**
+* **Application events for plugin configuration create and update**
```
Klaas-Ritense: release notes and documentation are too technical. Users dont need to know about internal events

documentation/release-notes/13.x.x/13.39.0/valtimo-backend-libraries.md
Klaas-Ritense, line 1: Files are not needed:
13.39.0/valtimo-backend-libraries.md
13.39.0/valtimo-frontend-libraries.md

backend/plugin/src/main/kotlin/com/ritense/plugin/service/PluginService.kt
```diff
@@ -148,24 +150,24 @@ class PluginService(
...
             pluginConfiguration.runAllPluginEvents(EventType.CREATE)
-
         } catch (e: Exception) {
             pluginConfigurationRepository.deleteById(pluginConfiguration.id)
             throw PluginEventInvocationException(pluginConfiguration, e)
         }
-
+        applicationEventPublisher.publishEvent(PluginConfigurationCreatedEvent(pluginConfiguration))
+        // TODO remove publish of PluginsDeployedEvent in next major version
         applicationEventPublisher.publishEvent(PluginsDeployedEvent())
```
Klaas-Ritense: Its possible to remove PluginsDeployedEvent. Its an internal event and not a breaking change.

Klaas-Ritense (approved):
There is a merge conflict

backend/plugin/src/main/kotlin/com/ritense/plugin/service/PluginService.kt
maarten-ritense: Decided to remove publishing PluginsDeployedEvent from plugin service.

PR #843, opened 2026-07-23
Making it possible to skip a timers from the progress tab

IvarKoreman-Ritense:
### Describe the changes
Link to the related Github issue: https://github.com/generiekzaakafhandelcomponent/gzac-issues/issues/465

### Breaking changes

### Documentation

- [x] Yes
- [ ] Not applicable

### Tests
- [x] Yes
- [ ] Not applicable

- [ ] Yes
- [x] Not applicable

Describe the testing steps
- [ ] Have a process that has a timer and start this process and walk through it until you're waiting on this timer
- [ ] Login as admin (who will have the OperatonExecution->modify PBAC permission)
- [ ] Go to the case of the above mentioned process and go to the Progress tab.
- [ ] Click on the skip icon above the active timer and skip it.
- [ ] Repeat the first steps and login as a user without the OperatonExecution->modify PBAC permission to ensure that they don't have this skip option.

### Security
- [x] Yes
- [ ] Not applicable

- [x] Yes
- [ ] Not applicable

- [x] Yes
- [ ] Not applicable

### Dependencies
- [ ] Yes
- [x] Not applicable


backend/process-document/src/main/kotlin/com/ritense/processdocument/web/rest/ProcessTimerResource.kt
```diff
@@ -0,0 +1,156 @@
...
+
+        return ResponseEntity.noContent().build()
+    }
+
+    /**
+     * Loads the process-instance-level execution and requires the [OperatonExecutionActionProvider.MODIFY]
+     * permission on it. The process-instance execution row has `ID_ == PROC_INST_ID_`, so it is loaded by
+     * [processInstanceId]. A missing execution means the instance is not active.
+     */
+    private fun requireExecutionModifyPermission(processInstanceId: String) {
```
ivo-ritense: I discussed this with the team, and we think it's best to have a new entity type for this. The risk right now is that we might in the future want to add other ways of changing a running process that would make sense to also put on operaton execution. Users might then get unintended access to new features. 

Our suggestion so far is to make an `OperatonTimer` entity and perform the `complete` action. That way we could later also have a different feature to change the timer to a different duration as the `modify` action.

It would make sense to add a mapper from the `OperatonTimer` to the `OperatonExecution` class so you can still refer to the path leading up to the case instances for conditions. 

As an example on how something like this could be done, you can look at the [ZgwDocument](https://github.com/valtimo-platform/valtimo/blob/ac286f2544295a35464bddb0a5135179769f9fa6/backend/zgw/documenten-api/src/main/kotlin/com/ritense/documentenapi/authorization/ZgwDocument.kt#L4) class. Used like this https://github.com/valtimo-platform/valtimo/blob/ac286f2544295a35464bddb0a5135179769f9fa6/backend/zgw/documenten-api/src/main/kotlin/com/ritense/documentenapi/client/DocumentenApiClient.kt#L82

ivo-ritense (approved):

frontend/projects/valtimo/case/src/lib/components/case-detail/tab/progress/progress.component.ts
```diff
@@ -17,8 +17,30 @@
...
+  tap,
+} from 'rxjs';
 import {ListItem} from 'carbon-components-angular/dropdown';
+import {CaseProcessTimerService} from '../../../../services';
+
+interface PendingSkip {
+  timer: SkippableTimer;
+  documentId: string;
+  processInstanceId: string;
+}
```
sofiaIvarsRitense: Please move his interface to a different file

frontend/projects/valtimo/process/src/lib/process-diagram/process-diagram.component.ts
```diff
@@ -112,12 +124,22 @@ export class ProcessDiagramComponent implements OnInit, OnDestroy, OnChanges {
     });
   }
 
-  ngOnChanges(): void {
-    if (this.processDefinitionKey) {
+  ngOnChanges(changes: SimpleChanges): void {
```
sofiaIvarsRitense:
```suggestion
  public ngOnChanges(changes: SimpleChanges): void {
```

frontend/projects/valtimo/process/src/lib/process-diagram/process-diagram.component.ts
```diff
@@ -112,12 +124,22 @@ export class ProcessDiagramComponent implements OnInit, OnDestroy, OnChanges {
...
+    ) {
       this.loadProcessInstanceXml(this.processInstanceId);
     }
+
+    if (changes['skippableTimers'] || changes['canSkipTimer']) {
+      this.renderSkipTimerOverlays();
+    }
   }
 
   ngOnDestroy() {
```
sofiaIvarsRitense:
```suggestion
  public ngOnDestroy() {
```

frontend/projects/valtimo/case/src/lib/components/case-detail/tab/progress/progress.component.ts
```diff
@@ -81,14 +121,104 @@ export class CaseDetailTabProgressComponent {
...
+
+  /**
+   * The case and process instance the timer belongs to are captured here, so confirming skips the
+   * timer in the context it was requested from, even when the selection or the route changed in the
+   * meantime.
+   */
+  public onRequestSkipTimer(timer: SkippableTimer): void {
+    const processInstanceId = this.selectedProcessInstanceId$.value;
+
+    this._documentId$.pipe(take(1)).subscribe(documentId => {
```
sofiaIvarsRitense: Could you add _takeUntilDestroyed_ or __subscriptions_ to manage to close this subscription, please?

frontend/projects/valtimo/process/src/lib/process-diagram/process-diagram.component.scss
```diff
@@ -55,3 +55,27 @@
...
   }
 }
+
+.valtimo-skip-timer-overlay {
+  align-items: center;
+  background-color: #0f62fe;
+  border: none;
+  border-radius: 50%;
+  box-shadow: 0 1px 3px rgba(0, 0, 0, 0.3);
+  color: #ffffff;
```
sofiaIvarsRitense: Please for the colors of this file, try to use Carbon tokens when it's possible ($button-primary, $button-primary-hover, $text-on-color, $focus)

IvarKoreman-Ritense: I let Claude process the feedback, their response for using `--cds-*` instead of `$button-primary` etc.:

> I used the CSS custom property form (var(--cds-*)) rather than the Sass $button-primary form. Valtimo's Carbon theme
  (components/assets/css/carbon.scss:102-106) maps $button-primary/$button-primary-hover/$focus onto
  --vcds-color-60/70, i.e. the runtime tenant accent color set by the theme service. Sass variables would be baked in at
  build time and lose both the accent color and light/dark switching. This is also the pattern the rest of the codebase
  uses (e.g. color-picker.component.scss, overflow-menu.component.scss).

My frond-end knowledge isn't good enough to determine if Claude is correct.

sofiaIvarsRitense (approved):

PR #847, opened 2026-07-23
fix: carbon list tracking, tooltip destroy and case/dossier translation consistency

mbritense:
-


marijnritense (approved):

frontend/projects/valtimo/components/src/lib/directives/tooltip/tooltip.directive.ts
```diff
@@ -66,4 +74,11 @@ export class TooltipDirective implements OnInit {
       this.overlayRef.detach();
     }
   }
+
+  ngOnDestroy(): void {
```
sofiaIvarsRitense: public

sofiaIvarsRitense (approved):

PR #848, opened 2026-07-23
More stable integration tests

Klaas-Ritense:
## Problem

The `release/13.38.0` publish pipeline [failed](https://github.com/valtimo-platform/valtimo/actions/runs/30002112640) on `:backend:zgw:catalogi-api:integrationTesting` with a `java.net.BindException`. The other ZGW jobs it reported were only *canceled* by matrix `fail-fast`.

Several integration tests start their `MockWebServer` on a **fixed port inside the OS ephemeral range** (Linux `32768–60999`, macOS `49152–65535`):

| Module | Port |
|---|---|
| catalogi-api | 56274 |
| zaken-api | 56273 |
| documenten-api | 56273 |
| keycloak-iam | 49152 |
| object-management | 47797 |

The kernel draws from that same range to auto-assign local ports to **outbound** sockets the app opens at context startup (e.g. HikariCP → Postgres). If it hands out `56273` before the test binds its listener, `server.start(port = …)` fails with `BindException`. It's a race — hence intermittent, and why it only occasionally breaks CI.

## Fix

Two approaches, because the modules differ:

- **catalogi-api** — no static config coupling, so fixed the right way: `server.start()` (OS-assigned free port) and derive the URL constants from `server.url(...)`, matching every other healthy ZGW test. Also removed a `Thread.sleep(2000)` band-aid.
- **keycloak-iam, object-management, zaken-api, documenten-api** — the port is hardcoded in **static Spring-loaded config** (`application.yml`, OIDC discovery JSON, `app.pluginconfig.json`), so the mock server must keep a fixed port. Moved each **below 32768** (`19152 / 17797 / 16273 / 16274`) so the OS never auto-assigns it to an outbound socket. Kotlin and resource files change together, so all URLs stay consistent — no logic change.

## Verification

`integrationTestingPostgresql` for all five modules: **98 tests, 0 failures.**

## Notes

- Test-only change; safe to cherry-pick into `release/13.38.0`.
- `documenten-api-preview` and `zaakdetails` also hardcode `56273` in their `app.pluginconfig.json`, but their tests don't do a fixed-port `.start()`, so they're not part of this `BindException` class. Left out of scope.
- Longer term, the static-config modules could migrate to fully dynamic ports (like catalogi-api) via `@DynamicPropertySource`.

### DIFF
--- a/backend/zgw/zaken-api/src/test/kotlin/com/ritense/zakenapi/ZakenApiPluginIT.kt
+++ b/backend/zgw/zaken-api/src/test/kotlin/com/ritense/zakenapi/ZakenApiPluginIT.kt
@@ -107,7 +107,7 @@ class ZakenApiPluginIT : BaseIntegrationTest() {
     internal fun setUp() {
         server = MockWebServer()
         setupMockZakenApiServer()
-        server.start(port = 56273)
+        server.start(port = 16273)
         sleep(2000) // Needed to fix connection refused error
 
         // Since we do not have an actual authentication plugin in this context we will mock one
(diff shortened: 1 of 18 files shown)


backend/zgw/zaken-api/src/test/kotlin/com/ritense/zakenapi/ZakenApiPluginIT.kt
```diff
@@ -107,7 +107,7 @@ class ZakenApiPluginIT : BaseIntegrationTest() {
     internal fun setUp() {
         server = MockWebServer()
         setupMockZakenApiServer()
-        server.start(port = 56273)
+        server.start(port = 16273)
```
ivo-ritense: Why can we not do the same thing here? Start with a random port, use it in all urls and remove the sleep?

Klaas-Ritense: Ive randomized all ports

ivo-ritense (approved):

PR #849, opened 2026-07-23
GH-824: Fixed open search sorting (previously not working) and default sorting preferences that weren't being applied correctly

marijnritense:
Solves https://github.com/generiekzaakafhandelcomponent/gzac-issues/issues/824


backend/case/src/main/kotlin/com/ritense/case/service/CaseInstanceService.kt
```diff
@@ -174,8 +175,11 @@ class CaseInstanceService(
 
     private fun mutatePageable(caseListColumns: Collection<CaseListColumn>, pageable: Pageable): PageRequest {
         val newSortOrders = pageable.sort.map { sortOrder ->
-            val caseListColumn = caseListColumns.find { caseListColumn -> caseListColumn.id.key == sortOrder.property }
+            val sortKey = sortOrder.property.removePrefix("$.")
+            logger.info { "mutatePageable: sortKey=$sortKey, columns=${caseListColumns.map { it.id.key }}" }
```
ivo-ritense: I don't think we should do multiple info log lines every time a search is done.

backend/case-opensearch/src/main/kotlin/com/ritense/document/opensearch/service/JsonSchemaDocumentOpenSearchService.kt
```diff
@@ -385,20 +391,35 @@ class JsonSchemaDocumentOpenSearchService(
...
+                queryBuilder.withSorts(sortBuilder)
+                logger.info { "executeSearch: adding sort $osField $sortOrder" }
+            }
+        }
 
+        val dataQuery = queryBuilder.build()
         val hits = elasticsearchOperations.search(dataQuery, JsonSchemaDocumentOsDocument::class.java)
         val total = hits.totalHits
         val ids: List<String> = hits.searchHits.mapNotNull { it.id }
+        logger.info { "executeSearch: returned ${ids.size} IDs in order: $ids" }
```
ivo-ritense: Again, this could maybe be trace, but i don't feel most of these log lines being useful

ivo-ritense (approved):

Klaas-Ritense (approved):

PR #857, opened 2026-07-28
Skip documents missing in the Documenten API when listing case documents

mbritense:
closes https://github.com/generiekzaakafhandelcomponent/gzac-issues/issues/799


backend/zgw/zaken-api/src/main/kotlin/com/ritense/zakenapi/service/ZaakDocumentService.kt
Klaas-Ritense, line 1: Right now its only a partial fix. Things like `ZaakDocumentService.deleteRelatedInformatieObjecten` will still break

ivo-ritense (approved):

Klaas-Ritense (approved):

PR #868, opened 2026-07-30
feat: building block hardening

mbritense:
- Building block call activities are validated on save, update, and at runtime: a missing, wrong, or namespace-shadowed business key mapping now fails with a clear fix-it message instead of the building block silently running against the wrong case data
- The validation understands engine namespace rules: a correct `camunda:in` that is ignored because `operaton:in` elements are present is detected and explained
- Deploying a new process version with a broken building block call activity logs an error pointing at the problem
- Input mappings must target building block fields and output mappings must read from them; invalid mappings are rejected on save instead of silently ignored
- Plugin action properties that resolve to no value are logged with property, activity, and process, making silent misbehaviour (e.g. mails without attachments) diagnosable
- Regression tests prove passing values and file references into a building block, resolving them in plugin actions, and prefilling task forms from the building block document
- Documentation added on how data and file attachments reach a building block and what the call activity must look like
- Release notes for 13.40.0 filled in


backend/building-block/src/main/kotlin/com/ritense/buildingblock/processlink/mapper/BuildingBlockProcessLinkMapper.kt
```diff
@@ -271,6 +281,36 @@ class BuildingBlockProcessLinkMapper(
...
+        inputMappings: List<BuildingBlockInputMapping>,
+        outputMappings: List<BuildingBlockOutputMapping>
+    ) {
+        val invalidInputTargets = inputMappings
+            .map { it.getPrefixedTarget() }
+            .filter { !it.startsWith(DOC_TARGET_PREFIX) }
+        require(invalidInputTargets.isEmpty()) {
+            "Input mapping targets must be building block fields. A value passed to a building block is " +
+                "stored in the building block document and can be referenced inside the building block " +
+                "with 'doc:'. Invalid targets: ${invalidInputTargets.joinToString()}"
```
marijnritense: That's a lot of words! I don't think this is a showstopper or anything, but if you have the chance, can you make this a bit less wordy? This applies to the other require as well

documentation/release-notes/13.x.x/13.40.0/README.md
```diff
@@ -12,10 +12,36 @@
...
+  When a plugin action property resolves to no value — for example an attachment list that references a field that
+  does not exist in the current context — a warning is now logged that names the property, the activity, and the
+  process definition. This makes it much easier to find out why a plugin action behaves as if a value was never
+  provided.
 
 ## Bugfixes
 
 * New bugfix.
+
+## Documentation
```
marijnritense: This isn't a section for the release notes. We link to relevant release notes on the individual release notes. I see the purpose of it, though, so maybe link it in the "Clearer rules for building block input and output mappings", or one of the other ones you added?

backend/plugin/src/main/kotlin/com/ritense/plugin/service/PluginService.kt
```diff
@@ -691,10 +693,36 @@ class PluginService(
...
     }
 
+    /**
+     * A property that resolves to null is passed to the plugin action as null, which typically makes
+     * the action silently skip the related behaviour (e.g. sending a mail without attachments). Log
+     * it, so misconfigured references are diagnosable. A common cause is referencing a process
+     * variable (pv:) inside a building block: values passed to a building block only exist in the
+     * building block document (doc:), never as process variables.
+     */
+    private fun logUnresolvedActionProperties(
```
ivo-ritense, line 712: I'm not sure i like this. i think there can be valid cases where some data is null. if every times that is logged as warn this might cause all kinds of spam. I would either remove this, or change the log level to debug.

documentation/release-notes/13.x.x/13.40.0/README.md
```diff
@@ -12,10 +12,36 @@
 
 ## Enhancements
 
-* **New enhancement title**
+* **Building block call activities are now validated**
 
-  New enhancement explanation.
+  The business key configuration of a building block call activity is checked when the building block link is saved
```
ivo-ritense: I feel that some of these release notes go too much into the technical detail. They should be easily understandable and preferably not too long.

ivo-ritense (approved):

ivo-ritense:
Validation works for business key and input mappings when saving a process, but the feedback is in the console instead of to the user.

PR #869, opened 2026-07-30
Export and import a system process with the elements it references

maarten-ritense:
### Describe the changes
Link to the related Github issue: https://github.com/generiekzaakafhandelcomponent/gzac-issues/issues/481

A process that is not linked to a case definition (the processes on `/processes`) could only be moved between environments as a bare BPMN file, so its process links — and everything it references — had to be reconnected or redeployed by hand on every next environment. This adds an export that bundles the process **with everything it needs**, and an import for that package.

The export now writes a complete package under `config/global/`: the process BPMN, its process links, and — walked recursively — the **called sub-processes, decision tables and forms** it references. This is the same closure a case-definition export produces. Importing it deploys the process and every bundled element; elements that already exist on the target are (re)deployed, and the import preview lists which ones **will be replaced** so the user consents first.

The import half already existed and was largely unreachable: `GlobalProcessDefinitionImporter`, `GlobalDecisionDefinitionImporter`, `GlobalFormDefinitionImporter`, `GlobalProcessLinkImporter` and `ValtimoImportService.importGlobal` already consume `config/global/{bpmn,dmn,form,process-link}/*` — that is how the dev app deploys `document-upload` and its plugin links at startup. The work is almost entirely on the **export** side (no exporter wrote called elements into `config/global`), plus the REST endpoints, the preview and the UI.

Specify the code branch location: `story/481-system-process-export-include-process-links` → `next-minor`

**Relevant comments:**
> **Scope: the package is complete (this changed during review).** An earlier version of this branch deliberately excluded called sub-processes, decision tables and forms and only *reported* them as missing on import. Reviewer feedback (marijnritense) was that this sets wrong expectations — a process that references other definitions should export them too. The agreed acceptance criterion is **parity with the case-definition export**, so the recursive walk is now on: a call activity pulls in its sub-process (and that sub-process's own called elements, transitively), a business rule task pulls in its decision table, and a form process link pulls in its form. **Form flow definitions are the one deliberate exclusion**: they are architecturally case/building-block scoped (the blueprint is part of the `form_flow_definition` primary key, there is no global storage or `deployGlobalFormFlow`), so a case-unlinked process cannot hold a resolvable form-flow link. Including them would need a schema migration plus a change to the shared `BlueprintOwner`/`BlueprintType` contract; out of scope, raise separately if ever needed.
>
> **The import preview now shows what will be *replaced*, not what is *missing*.** Because the package now carries the referenced elements, importing can update an element that already exists on the target and may be shared with other processes. The preview therefore returns `elementsToReplace` (processes, decision tables, forms that already exist here and will be overwritten), surfaced in the review step as a warning so replacing them is a conscious choice; the import stays enabled. The old "missing reference" list narrows to references that genuinely **could not be bundled**.
>
> **Resolvable-but-missing fails loudly; unbundlable is skipped and reported.** A `deployment` binding or an expression based `calledElement` / `decisionRef` (`${…}` / `#{…}`) cannot be resolved to a specific definition, so it is left out of the package and named by the import preview as "could not be included". A reference that *can* be resolved statically (a literal key with a latest/version/versionTag binding) but points at a definition that is **not deployed here** now **fails the export** with a descriptive message, matching `ProcessDefinitionExporter` — bundling something that references a missing definition would produce a broken package. A missing form referenced by a process link fails the same way; because the form is referenced by id, the message names the **activity and process** the link belongs to (the form's name is gone with the form), or the form name when it is still resolvable. The frontend shows a dedicated export-error toast for this instead of the misleading "Process failed to deploy".
>
> **⚠ Behaviour change: `config/global` process links are leading.** The imported process-link file is the complete set of links for a process, so links not in the file are removed. This applies to importing a package *and* to the startup autodeployment of `config/global/process-link/*.json`. A link added through the interface on top of a process that also ships a `config/global` process-link file disappears on the next restart. Chosen deliberately (configuration stays authoritative, environments cannot drift) and described in the release notes.
>
> **Read-only system processes are protected.** Importing over a process that exists as a system process which may not be updated (`valtimo.process.systemProcessUpdatable=false`) is refused in the REST layer; `GlobalProcessDefinitionImporter` keeps `skipIsDeployableCheck = true` so startup autodeployment retains authority over its own system processes.
>
> **Import preview only considers global process links (review follow-up).** `ProcessDefinitionImportPreviewService` and `PluginConfigurationImportPreviewContributor` matched any `.../process-link/*.process-link.json` in the archive, so case- and building-block-scoped process links in an uploaded package were treated as the process's own global links — corrupting the missing-reference / element-to-replace preview and the plugin mapping. The regex is now anchored on `global/`, matching the global exporter/importer path (resolves marijnritense's comment); covered by a new case-scoped regression test.
>
> **Export manifest.** A system process export produces a manifest: `ArtifactType.PROCESS_DEFINITION`, with the plugins of its process links as dependencies. A bpmn has no field a manifest can `$ref` into, so both values are literals: title = process name (falling back to key), version = the `operaton:versionTag` (stable across environments), falling back to the deployment version. A version tag that encodes a case or building block definition is ignored. This needed a 2-line change to `ValtimoExportService`: it collected `manifestDependencies` only from non-root exporters, which assumed one exporter per root request; the bpmn and the process links are two exporters of the same request, so dependencies are now collected from any result that does not itself contribute the root artifact. No behaviour change for existing exports.
>
> **UI: the export actions were relabelled and reordered.** The process menu now offers **Export** (the full package — process + links + referenced elements) as the primary action, and **Export BPMN definition** for the bare BPMN. Previously "Export" was the bare BPMN and "Export with process links" the package. During the import, plugin links in the package can be pointed at the target environment's plugin configuration via a new standalone plugin-configuration-mapping component in `@valtimo/plugin`.
>
> **Refactor: `ProcessLinkResource` was split by responsibility.** The single 22-endpoint controller became three focused controllers serving the same URLs — `ProcessLinkResource` (base `/api/v1/process-link` CRUD), `CaseProcessDefinitionManagementResource` (case-definition process management) and `ProcessDefinitionManagementResource` (case-unlinked process management incl. export/import/preview/validate) — sharing a `ProcessDefinitionResponseAssembler`. The security configurer was split to match (one per surface, plus one for the runtime task endpoints), and the unit + integration tests were split per resource. URLs, request/response contracts and authorization rules are unchanged. A dead endpoint (`getSingleProcessDefinitionWithLinks` — unreachable behind `denyAll` and unused by the frontend) and an orphan security matcher (no controller mapped it) were removed along the way.
>
> **Docs target 13.43.0.** 13.42.0 releases first with other teams' notes, so these release notes were retargeted to 13.43.0.

### Breaking changes

No schema migration (form flows were deliberately excluded to avoid one). Packages exported by the earlier version of this branch (bpmn + process-link only) still import — the added importers simply find no dmn/form/extra-bpmn entries. The API URLs and contracts are unchanged by the controller split.

### Documentation

`documentation/release-notes/13.x.x/13.43.0/README.md` and the system-process configuration guide at `documentation/configuration-guides/system-processes/` (both added to `SUMMARY.md`). The `next-minor` merge brought in the new documentation standard, which removed the old `documentation/features/process/` and `documentation/using-valtimo/process/` trees, so the system-process docs were re-authored in the new structure.

- [x] Yes

### Tests
- [x] Yes

`GlobalProcessDefinitionExporterTest` (walk, building-block/deployment/expression skip, fail-on-missing), `GlobalDecisionDefinitionExporterTest`, `GlobalFormDefinitionExporterTest` (incl. clear missing-form message), `GlobalProcessLinkExporterTest`, `FormProcessLinkMapperTest`, `ProcessDefinitionImportPreviewServiceTest` (replace list, unresolved refs, and case-scoped process links ignored), `PluginConfigurationImportPreviewContributorTest` (new), `GlobalProcessLinkImporterTest`, the split resource tests, and the frontend upload/builder specs.

- [x] Yes

`GlobalProcessDefinitionExporterIntTest` (referenced DMN bundled, deployed with the process or separately), `GlobalProcessDefinitionRoundTripIntTest`, `GlobalProcessLinkImporterIntTest`, `ProcessDefinitionImportPreviewServiceIntTest` (incl. `elementsToReplace`), `GlobalProcessDefinitionExportManifestIntTest`, and the split `ProcessLinkResourceIT` / `CaseProcessDefinitionManagementResourceIT`.

Describe the testing steps
- [ ] `Export` on `/processes/document-upload` downloads `<key>_v<version>_<timestamp>.process.zip` containing, under `config/global/`, the root bpmn, every called sub-process bpmn + its process-link json, every referenced dmn, every form referenced by a process link (recursively), and `valtimo-manifest.json`
- [ ] `Export BPMN definition` on the same process downloads only the bare `<key>.bpmn`
- [ ] Importing that package on a clean environment deploys the process and every referenced element; all process links resolve and point at the mapped plugin configurations
- [ ] Importing over an environment that already has some of those elements lists exactly those under "will be replaced" in the review step, and importing replaces them
- [ ] Exporting a process whose business rule task references a decision table (or call activity a sub-process) that is **not deployed** here fails with the export-specific error toast naming the missing element, instead of a package with a dangling reference
- [ ] Importing over a protected system process is refused with "managed by the application configuration", with `Upload` disabled
- [ ] Uploading a bare `.bpmn` behaves exactly as before
- [ ] The file dialog filters on `.bpmn` / `.zip` in Firefox and Safari — cannot be automated, `setInputFiles` bypasses the dialog

### Security
- [x] Yes

The import refuses a package before applying any of it when it would fail or when it would overwrite a process managed by configuration, so no partially imported state can result. The preview and the import both derive their decisions from the archive contents server-side; nothing from the client decides what may be replaced. The export fails before producing a package that references a definition that is not present. A dead, unreachable endpoint and an orphan security matcher were removed, reducing surface.

- [x] Yes

The management endpoints (now split across `ProcessLinkResource`, `CaseProcessDefinitionManagementResource` and `ProcessDefinitionManagementResource`) are restricted to `ADMIN`; the runtime task endpoints require an authenticated user. This is enforced by four ordered `HttpSecurityConfigurer`s (`ProcessLinkHttpSecurityConfigurer`, `ProcessLinkTaskHttpSecurityConfigurer`, `CaseProcessDefinitionManagementHttpSecurityConfigurer`, `ProcessDefinitionManagementHttpSecurityConfigurer`), all ordered below the `denyAll` fallback, and covered by `ProcessLinkSecuritySmokeIntegrationTest`. URLs are unchanged by the split.

- [ ] Yes
- [x] Not applicable

These are ADMIN-only management endpoints that run under `runWithoutAuthorization`, exactly like the existing case-definition export and import. There is no per-resource permission model for process definitions to check against.

### Dependencies
- [x] Yes

No new dependencies were added.


frontend/projects/valtimo/plugin/src/lib/components/plugin-configuration-mapping/plugin-configuration-mapping.component.ts
```diff
@@ -0,0 +1,232 @@
...
+import {ChangeDetectionStrategy, Component, Input, OnInit} from '@angular/core';
+import {FormBuilder, FormGroup, ReactiveFormsModule} from '@angular/forms';
+import {TranslateModule, TranslateService} from '@ngx-translate/core';
+import {ComboBoxModule, LayerModule, ListItem, NotificationModule} from 'carbon-components-angular';
+import {BehaviorSubject, forkJoin, Observable, take} from 'rxjs';
+import {PluginConfiguration} from '../../models';
+import {PluginManagementService} from '../../services/plugin-management.service';
+import {PluginTranslationService} from '../../services/plugin-translation.service';
+
+type PluginMappingStatus = 'available' | 'no-configurations' | 'not-installed';
```
sofiaIvarsRitense: Please move this type to a different file

frontend/projects/valtimo/plugin/src/lib/components/plugin-configuration-mapping/plugin-configuration-mapping.component.ts
```diff
@@ -0,0 +1,232 @@
...
+
+interface PluginConfigurationPreview {
```
sofiaIvarsRitense: Please move this interface to a different file

frontend/projects/valtimo/plugin/src/lib/components/plugin-configuration-mapping/plugin-configuration-mapping.component.ts
```diff
@@ -0,0 +1,232 @@
...
+  pluginConfigurationId: string;
+  pluginDefinitionKey: string | null;
+  existsInTargetEnvironment: boolean;
+}
+
+interface PluginMappingRow {
```
sofiaIvarsRitense: Please move this interface to a different file

sofiaIvarsRitense (approved):

backend/core/src/main/kotlin/com/ritense/valtimo/exporter/GlobalProcessDefinitionExporter.kt
```diff
@@ -0,0 +1,94 @@
...
+import com.ritense.valtimo.operaton.domain.OperatonProcessDefinition
+import com.ritense.valtimo.operaton.service.OperatonRepositoryService
+import org.operaton.bpm.engine.RepositoryService
+import org.operaton.bpm.model.bpmn.Bpmn
+import java.io.ByteArrayOutputStream
+
+/**
+ * Exports the BPMN of a process definition that is not part of a case definition.
+ *
+ * Unlike [ProcessDefinitionExporter] this exporter does not create related export requests for
```
marijnritense, line 35: Why do we not want to export called subprocesses/decision definitions?

backend/form-flow/src/main/kotlin/com/ritense/formflow/mapper/FormFlowProcessLinkMapper.kt
```diff
@@ -193,6 +195,26 @@ class FormFlowProcessLinkMapper(
 
     override fun getImporterType() = "formflow"
 
+    override fun getMissingReference(deployDto: ProcessLinkDeployDto, blueprintId: BlueprintId?): MissingReferenceDto? {
+        deployDto as FormFlowProcessLinkDeployDto
+        val definition = when (blueprintId) {
```
marijnritense, line 200: You're calling the same method twice, why are you using a when for this?

backend/core/src/main/kotlin/com/ritense/valtimo/exporter/GlobalProcessDefinitionExporter.kt
marijnritense: Discussed it a bit more internally, we would like to see this included before we accept it into the product. People will get wrong expectations if they export a process definition that references other process or decision definitions.

maarten-ritense: The idea is/was to mimic the existing export of a 'system' process with its process links included. As the existing system process export only exports the bpmn definition and not also its called elements I made the decision for this new export to exclude these as well.

Including these will extend the scope significantly and introduces more complexity if you ask me. It will do a lot more than the existing 'simple' system process export. What should the scope be then, include all called bpmn elements including and referenced form definitions too? What if called elements are used by other bpmn's as well, same question for forms.

Maybe we should discuss this a little more or it could be an iteration of this in the (near) future if there is demand for this.

backend/process-link/src/main/kotlin/com/ritense/processlink/service/ProcessDefinitionImportPreviewService.kt
```diff
@@ -0,0 +1,322 @@
...
+    }
+
+    companion object {
+        private val logger = KotlinLogging.logger {}
+        private const val PROCESS_DEFINITION_ID = "processDefinitionId"
+        private const val PROCESS_DEFINITION_ID_PLACEHOLDER = "-"
+        private val BPMN_REGEX = """.*/?global/bpmn/(?:.*/)?(.+)\.bpmn""".toRegex()
+        private val DMN_REGEX = """.*/?global/dmn/(?:.*/)?(.+)\.dmn""".toRegex()
+        private val FORM_REGEX = """.*/?global/form/(?:.*/)?(.+)\.form\.json""".toRegex()
+        private val PROCESS_LINK_REGEX = """.*/?process-link/(?:.*/)?(.+)\.process-link\.json""".toRegex()
```
marijnritense: I think this also matches case process links. Can you add the global part to the regex?

documentation/configuration-guides/system-processes/README.md
```diff
@@ -0,0 +1,112 @@
+# System processes
+
+System processes are [processes](../../fundamentals/process.md) that are critical to the functioning of Valtimo itself. Because they must be handled with care, they are marked as read-only by default: a read-only process can be used, but it cannot be edited, saved, or overwritten by an import.
+
+System processes are managed from **Admin** > **Processes**, where they are listed with the status **Read-only**.
```
marijnritense: This reads different from the other configuration guides. Should be a stepper with 2 steps, and include a screenshot showing how to get there.

documentation/configuration-guides/system-processes/README.md
```diff
@@ -0,0 +1,112 @@
...
+
+{% hint style="warning" %}
+A read-only system process can be used but not modified. To change one, make system processes updatable first.
+{% endhint %}
+
+---
+
+## Making a system process updatable
+
+Whether system processes are read-only is controlled by a back-end setting. By default it is off, so system processes are read-only. To allow them to be changed, set the following property in the `application.yml` of the back-end implementation:
```
marijnritense: This is too technical, this entire section can be removed

documentation/configuration-guides/system-processes/README.md
```diff
@@ -0,0 +1,112 @@
...
+While a process is read-only, the following is not possible:
+
+- Editing the diagram or saving (deploying) changes
+- Migrating running process instances to it
+- Overwriting it by importing a package (see [Import and export](#import-and-export))
+
+<figure><img src="../../assets/configuration-guides/system-processes/04-read-only-process-editor.png" alt=""><figcaption>A system process in the read-only viewer</figcaption></figure>
+
+{% hint style="warning" %}
+A read-only system process can be used but not modified. To change one, make system processes updatable first.
```
marijnritense: There's no UI way to configure this, so I think the best way to handle this is to not mention it right now.

documentation/configuration-guides/system-processes/README.md
```diff
@@ -0,0 +1,112 @@
...
+
+## Read-only behavior
+
+When a process is a system process, the process editor shows a red **System process** tag and a blue **Read-only** tag, and the diagram opens in a viewer that cannot be edited. The **Save** button is disabled.
+
+<figure><img src="../../assets/configuration-guides/system-processes/03-read-only-system-process.png" alt=""><figcaption>Read-only and System process tags</figcaption></figure>
+
+While a process is read-only, the following is not possible:
+
+- Editing the diagram or saving (deploying) changes
```
marijnritense: Can you make these things into a single hint? That draws extra attention to it. It should probably be moved down to just after the figure.

documentation/configuration-guides/system-processes/README.md
```diff
@@ -0,0 +1,112 @@
...
+
+---
+
+## Import and export
+
+{% hint style="success" %}
+Exporting and importing a system process as a full package is available since Valtimo `13.43.0`. Before that, only the BPMN definition could be exported.
+{% endhint %}
+
+A system process can always be exported. Use **Export** in the menu of the process to download a package.
```
marijnritense: Importing and exporting should be separate sections, with their own steps describing how to do this.

documentation/configuration-guides/system-processes/README.md
```diff
@@ -0,0 +1,112 @@
...
+<figure><img src="../../assets/configuration-guides/system-processes/05-upload-process-modal.png" alt=""><figcaption>Upload process definition</figcaption></figure>
+{% endstep %}
+{% step %}
+Click **Upload** to open the import preview.
+{% endstep %}
+{% endstepper %}
+
+A package cannot overwrite a process that exists on this environment as a read-only system process. In that case the import preview reports **Process is managed by configuration** and the import is blocked. Make the system process updatable first if the import has to replace it.
+
+### Mapping plugin configurations
```
marijnritense: This can be a part of the import section. But really, you can mostly use a similar text to what's in `configuration-guides/cases/README.md`'s "Uploading a case" section. This goes a bit too much into detail.

documentation/release-notes/13.x.x/13.43.0/README.md
```diff
@@ -4,18 +4,72 @@ Release date: 26-08-2026
 
 ---
 
+## New Features
```
marijnritense: Feels like there's way too much in the release notes here. Just document the feature, no need for all the tiny details as separate enhancements. In general there's too much text here, and unfortunately that means nobody will read the descriptions for these notes.

marijnritense (approved):

PR #882, opened 2026-08-05
fix: divider widget without a title no longer falls back to `-`

Rick-Ritense:
### Describe the changes
Link to the related Github issue: https://github.com/generiekzaakafhandelcomponent/gzac-issues/issues/843

Specify the code branch location: `next-minor`

**Relevant comments:**
> A divider widget configured without a title showed a `-` in two places: as the title in the
> widget list on the widget management page, and as the divider heading on the widget tab of a
> case. Both now stay empty. The value was never *stored* as `-` — it is `""` in the database in
> both cases, this was display only.
>
> The widget list `-` did not come from the widget code but from the shared `StringTypeConverter`,
> which renders `-` for any empty value in **every** carbon list. That placeholder is intentional
> in the rest of the admin screens, so rather than removing it globally it can now be overridden
> per column via `ColumnConfig.emptyPlaceholder`; the widget list opts out for the title column
> only. Default behaviour for every other list is unchanged.
>
> While testing I found a second, related problem: `Widget` required a non-blank title for every
> widget type. Case widget tabs use `CaseWidgetTabWidget` (no such check), which is why the
> reported scenario saved fine, but IKO views use `Widget` — so saving a divider without a title
> on an IKO view failed with a 500. A divider does not need a title, so the check now skips
> `DividerWidget`; all other widget types still require one.
>
> Note: with an empty title the divider element keeps its spacing (48px), so the visual separation
> between widget groups is preserved — it is an empty heading, not a removed one.

### Breaking changes

### Documentation

- [x] Not applicable

### Tests
- [x] Yes

`DividerWidgetTest` (back end), a divider case in `IkoWidgetManagementResourceTest`,
`stringTypeConverter.spec.ts` and `view-content.service.spec.ts` (front end). Both back-end tests
fail without the `Widget.kt` change, so they cover the regression.

- [x] Not applicable

Describe the testing steps
- [x] Step 1: Go to a widget admin page (Admin > Dossiers > *case* > Case details > Tabs > Widgets) and add a divider, filling in only the key and leaving the title empty. The Title column of the new row stays empty instead of showing `-`.
- [x] Step 2: Open a case of that case type and go to the widget tab. The divider renders without a title, and the spacing between the widget groups is preserved.
- [x] Step 3: Repeat step 1 on an IKO view (Beelden) — saving a divider without a title now succeeds instead of failing.

Verified manually against a locally running dev stack (back end + front end) in addition to the unit tests.

### Security
The Secure by Design principle has been applied to these changes
- [x] Not applicable

- [x] Not applicable

- [x] Not applicable

### Dependencies
- [x] Not applicable


sofiaIvarsRitense (approved):

ivo-ritense (approved):

documentation/release-notes/13.x.x/13.41.0/README.md
```diff
@@ -18,6 +18,19 @@
...
+  A divider widget that is configured without a title now stays empty, both in the widget list on the
+  widget management page and on the widget tab of a case. Previously a `-` was shown in both places as a
+  placeholder for the missing title. In addition, saving a divider without a title on an IKO view no longer
+  fails: the back end required a non-blank title for every widget, while a divider does not need one.
+
+* **A divider widget can be duplicated again**
+
+  Duplicating a divider was impossible because it opened the duplication modal with an empty (invalid) key
+  and no way to edit this key. This resulted in a disabled *Duplicate* button. The divider key is now always
+  filled with a default unique value, and it is possible to edit the key before duplicating it.
```
sofiaIvarsRitense:
```suggestion
  Fixed an issue where duplicating a divider opened the duplication dialog with an empty, invalid key that could not be edited, leaving the Duplicate button disabled. The dialog now pre-populates the divider key with a unique default value and allows it to be edited before duplicating.
```

PR #883, opened 2026-08-06
feat: form flow ui editor

mbritense:
closes https://github.com/orgs/generiekzaakafhandelcomponent/projects/5/views/3?sliceBy%5Bvalue%5D=WEEK+32&filterQuery=&pane=issue&itemId=224354789&issue=generiekzaakafhandelcomponent%7Cgzac-issues%7C844


sofiaIvarsRitense (approved):

marijnritense (approved):

backend/form-flow/src/main/kotlin/com/ritense/formflow/service/FormFlowRegistryService.kt
marijnritense, line 1: Just adding it as a comment here so you can respond. I don't think these are all things that need to happen, they're just things I'm noticing. I'll test more tomorrow.

1. When creating the form flow, I get an error the start step does not yet exist. Since every form flow needs one, it makes sense to create the form flow with the first step.
2. When the start step has been defined, I do not see an error (apart from a small ! at step-1, which also seems a bit blurry). Since I cannot click save and do not see any error, I don't know what is wrong.

  Image

3. When switching to the JSON editor, I would expect the step I just added (but couldn't save) to just show up correctly there, but with some fields not yet filled in.
4. Can we also have validation for the JSON?
5. Can we validate if any steps are unreachable?
6. And if there's loops that can't be broken out of?
7. I already mentioned this somewhere else as well, but it would be nice if we can split the expressions (for actions) into 1) selecting the bean, and then 2) selecting the method, and 3) filling in the arguments.
8. As an enhancement, I think it would be nice to be able to visualize the flow. 
9. Expressions in conditions would also benefit from additional help. Knowing what "basic" fields are present, maybe based on the forms added we can show a full list of fields available for each step even. This obviously has moments where it can break, so it's important to then also add feedback about that.

mbritense:
@marijnritense please re-review, I only picked up your points 1 and 2, which were bugs

mbritense:
3. Is not possible, only configurations able to be saved in BE shows up in JSON editor, if no changes saved yet, the user is warned (the state of each tab is independent)

marijnritense (approved):

documentation/release-notes/13.x.x/13.44.0/README.md
```diff
@@ -25,3 +32,4 @@ New enhancement explanation.
 | Area | Fix |
 |------|-----|
 | Cases | The process selector on the Progress tab shows long process names in full instead of cutting them off |
+| Process links | Changing the form flow definition on an existing form flow process link is now saved (previously the change was silently ignored). |
```
marijnritense:
```suggestion
| Process links | Changing the form flow definition on an existing form flow process link is now saved. |
```

marijnritense:
Found a bug with testing. When saving a form also adds a new "start step" step. 
image

It's in the database too so I'm not sure if this is a problem with not deleting the initial start-step, or if it actually gets added every time.

marijnritense (approved):

PR #581, opened 2026-04-20
Playwright test coverage for the Cases feature

sofiaIvarsRitense:
### Describe the changes
Link to the related Github issue: 
https://github.com/generiekzaakafhandelcomponent/atlas-internal/issues/304
https://github.com/generiekzaakafhandelcomponent/atlas-internal/issues/300
https://github.com/generiekzaakafhandelcomponent/atlas-internal/issues/303
https://github.com/generiekzaakafhandelcomponent/atlas-internal/issues/305

Specify the code branch location: playwright/case-playwright-tests

**Relevant comments:**
> 

### Breaking changes

- [x] The contribution only contains changes that are not breaking.

### Documentation

- [ ] Release notes have been written for these changes.

New features or changes that have been introduced have been documented.
- [ ] Yes
- [x] Not applicable

### Tests
Unit tests have been added that cover these changes
- [ ] Yes
- [x] Not applicable

Integration tests have been added that cover these changes
- [ ] Yes
- [x] Not applicable

Describe the testing steps
Please run the Playwright tests created for:
- case-details-management-forms
- case-details-management-decissions
- case-details-management-processes
- case-details-management-document

### Security
The [Secure by Design principle](https://en.wikipedia.org/wiki/Secure_by_design) has been applied to these changes
- [ ] Yes
- [x] Not applicable

Added or changed REST API endpoints have authentication and authorization in place
- [ ] Yes
- [x] Not applicable

Valtimo access control checks have been implemented
- [ ] Yes
- [x] Not applicable

### Dependencies
Newly added dependencies do not introduce known vulnerabilities/CVE's and are in line with the Valtimo license
- [ ] Yes
- [x] Not applicable

PR #895, opened 2026-08-12
GH-581: Check object permissions before calling the Objecten API

Klaas-Ritense:
https://github.com/generiekzaakafhandelcomponent/atlas-internal/issues/581

A user without permission to view objects was refused only after the object had already been
retrieved from the Objecten API. Because the API's answer came back first, the error such a user
received revealed whether the object existed — a one-bit existence oracle — and every request
that was going to be denied still made a pointless outbound call. The permission check now runs
before anything is requested, in all four read operations. The write operations already checked
first and are unchanged.

Nothing is lost by checking first: the entity passed to the check is a fieldless `Object()`, so
the check has no dependency on the response.

The four new tests each fail against the unpatched code with
`HttpClientErrorException$NotFound` — that failure *is* the leak, which is why the mock enqueues a
404 rather than a success.

## Scope

This fixes the concrete defect in issue 581 only. The issue also raises **per-object scoping**,
which is a permission-model question rather than a bug: `Object` is deliberately a fieldless
capability marker, so no permission condition can scope it. That is being handled as a roadmap
item and the issue should **not** be closed as fixed on the back of this PR — a `view` grant still
means "may read any object in a configured Objecten API".

Worth noting for the reporter: the write paths cited in the issue (`createObject`, `objectPatch`,
`objectUpdate`, `deleteObject`) already checked permission before the outbound call, so the
write/delete escalation described there does not apply to this defect.

Not verified against a running application: this is a reordering inside the API client with no UI
surface of its own, covered by unit tests that assert no request reaches the mock API when
permission is denied.

Note for the reviewer: this and #580 both add to `documentation/release-notes/13.x.x/13.41.0/README.md`
and will conflict there on merge. The conflict is additive — keep both entries.

### DIFF
--- a/documentation/release-notes/13.x.x/13.42.0/README.md
+++ b/documentation/release-notes/13.x.x/13.42.0/README.md
@@ -18,4 +18,8 @@
 
 ## Bugfixes
 
-* New bugfix.
+* **Object permissions are checked before the object is retrieved**
+
+  A user without permission to view objects is now refused before anything is requested from the Objecten API.
+  Previously the object was retrieved first, so the answer of the Objecten API could tell such a user whether an
+  object exists.
--- a/backend/zgw/objecten-api/src/main/kotlin/com/ritense/objectenapi/client/ObjectenApiClient.kt
+++ b/backend/zgw/objecten-api/src/main/kotlin/com/ritense/objectenapi/client/ObjectenApiClient.kt
@@ -51,12 +51,6 @@ class ObjectenApiClient(
         authentication: ObjectenApiAuthentication,
         objectUrl: URI
     ): ObjectWrapper {
-        val result = buildRestClient(authentication)
-            .get()
-            .uri(objectUrl)
-            .retrieve()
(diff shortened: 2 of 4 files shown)


documentation/release-notes/13.x.x/13.42.0/README.md
```diff
@@ -18,7 +18,11 @@
 
 ## Bugfixes
 
-* New bugfix.
+* **Object permissions are checked before the object is retrieved**
+
+  A user without permission to view objects is now refused before anything is requested from the Objecten API.
```
marijnritense, line 23: Can you also explicitly mention this means users have to configure access control for this now, or is this a non-issue since it's really just about the sharing of additional information?

Klaas-Ritense: This PR is a pure reorder. There are no changes to permissions needed

marijnritense (approved):

PR #912, opened 2026-08-14
Register Notificaties API abonnementen after startup, with exponential backoff

Klaas-Ritense:
Fixes the GZAC backend crash loop seen on Amsterdam Woningtoewijzen production and on the demo cluster (13.38.0), reported via support ticket 39420.

EC: https://github.com/generiekzaakafhandelcomponent/atlas-internal/issues/673

Refs [generiekzaakafhandelcomponent/gzac-issues#712](https://github.com/generiekzaakafhandelcomponent/gzac-issues/issues/712), follow-up to #778.

## The race

`ApplicationStateService` publishes `ApplicationFullyReadyEvent` **synchronously**. Two listeners hang off it with no `@Order` between them:

| Listener | Effect |
|---|---|
| `BootstrapHealthIndicator.onApplicationFullyReady()` | Flips the `readiness` / `startup` health groups to UP |
| `PluginsDeployedEventListener.handleApplicationFullyReadyEvent()` | Registers abonnementen → Open Notificaties calls back |

Open Notificaties validates a new abonnement by immediately calling back to the configured `callbackUrl`. On Kubernetes that callback cannot succeed until kubelet has observed the readiness probe and the EndpointSlice has been updated — which only starts happening because of this very event. Registration therefore fired in the same synchronous dispatch that first made the pod *eligible* to be considered ready. Routing could not possibly exist yet.

Ordering the two listeners does not help. `retry(times = 3)` had no delay between attempts, so all three burned within milliseconds, well inside the probe window (chart default `periodSeconds: 20`). The listener then threw `NotificatiesApiAbonnementException`, taking the application down — into a restart loop.

It presented as intermittent only because a rolling update usually leaves the old pod alive to field the callback.

## The change

- **Off the event thread.** Registration is handed to a dedicated single-threaded executor. `applicationFullyReady = true` stays on the event thread, so the `PluginConfiguration*` / `ProcessLink*` guards are unaffected.
- **Convergent backoff.** `registerAbonnementenWithBackoff()` retries the whole registration: 2s, doubling, capped at 30s, under a 15m deadline. It stops when it *succeeds* — the deadline is only a backstop so a genuinely misconfigured `callbackUrl` does not retry forever. It matches the chart's `startupProbe` budget (90 × 10s).
- **The pod stays up.** Registration failure no longer throws out of the listener. Every attempt is logged, and giving up logs at ERROR.
- Retrying is safe: `registerAbonnementenForPluginNotificatiesApiPlugins()` already reconciles against `getAbonnementen()` and the link repository, and skips when `callbackUrl` and `kanalen` are unchanged.

The **runtime** paths — plugin configuration created/updated/deleted, process link changes — keep their existing synchronous three-attempt-then-throw behaviour. Those fire while the application is already serving traffic, where a failure is a genuine configuration error an admin should see immediately, not a routing race.

The registration executor is deliberately separate from `notificatiesApiTaskExecutor`: the retry loop sleeps between attempts and would otherwise hold a thread that inbound event processing needs. Its threads are daemons and are not awaited on shutdown, so a pending retry never delays JVM exit.

## Configuration

New, alongside the existing `valtimo.zgw.register-abonnementen`:

| Property | Default |
|---|---|
| `valtimo.zgw.abonnement-registration.initial-backoff` | `2s` |
| `valtimo.zgw.abonnement-registration.max-backoff` | `30s` |
| `valtimo.zgw.abonnement-registration.max-duration` | `15m` |

## Deliberately not included

The advisory `GET {origin(callbackUrl)}/api/v1/ping` pre-flight and an abonnement-state health indicator, both from the design doc. The backoff resolves the reported failure on its own; the pre-flight only shortens the window and depends on OVH load balancer hairpin, which is unverified. Easy to add later if it proves useful.

## Not addressed here — DevOps side

Independent of this PR, worth raising with the chart owners: the chart's default `readinessProbe` targets `GET :8080/api/v1/ping`, not `:8081/actuator/health/readiness`, so the `BootstrapHealthIndicator` wiring from #778 is bypassed entirely. `/api/v1/ping` is `permitAll()` and answers as soon as the servlet container is up, potentially before migrations and autodeployments finish. Shortening `periodSeconds` on that probe does not buy what it appears to.

## Testing

`./gradlew :backend:zgw:notificaties-api:check` on a clean build — 65 tests, 0 failures. Three new unit tests cover the handoff to the executor, retrying to the deadline without throwing, and stopping as soon as registration succeeds. Not verified against a live cluster; the race itself is not reproducible locally.


ivo-ritense:
Has to be updated with new documentation skill

ivo-ritense (approved):

PR #919, opened 2026-08-17
Include form definitions used by widgets in the case export

Klaas-Ritense:
EC: https://github.com/generiekzaakafhandelcomponent/atlas-internal/issues/674

Exporting a case did not include the forms that are shown by form widgets on a widget tab. The exported case file could
therefore not be imported: the import stopped with the message that the form definition could not be found.

Those forms are now part of the export. When a widget points to a form that no longer exists, the export reports which
form is missing instead of failing without explanation.

Cases that were exported before this change still miss the form and need to be exported again.


ivo-ritense (approved):

ivo-ritense:
Wouldn't it be better to export all the forms that are created for the case definition. I think it's an old artifact from version 12 that we have to find the relevant form by usage.

ivo-ritense:
Has to be updated with new documentation skill

PR #954, opened 2026-08-25
Fix three issues in the admin editors

Klaas-Ritense:
## Bug

Dropdown lists in the admin screens opened on top of their own search box instead of
below it. Separately, field pickers could come up empty when more than one was on the
same screen, and picking a building block version for a process link offered only the
first page of versions.

## Fix

Menus now open in the right place, every field picker fills, and every version of a
building block is offered.

Not verified against a running application — none of the three files have existing unit
tests, so this is covered by type checks only.


sofiaIvarsRitense (approved):

sofiaIvarsRitense:
I can only see a maximum of 5 versions available for the BB

https://github.com/user-attachments/assets/b3ab1496-0e05-480d-b46e-fbdea270d0e1

sofiaIvarsRitense:
I still can’t see the full list of versions

https://github.com/user-attachments/assets/5026dd1b-93e3-4c0d-a3be-b56275bb5747

ivo-ritense (approved):

Klaas-Ritense:
AI: automatic reply, generated and posted without human review.

Yes, and it is fixed in the version that was merged — there was a second cause on top of the `all=true` request fix. Carbon flips a combo box upward as soon as its menu would reach past a scrollable parent, which inside the process link modal is the modal content, so the version list opened over its own field and ran off the top of the modal; and macOS keeps its overlay scrollbar invisible until something is already scrolling, so nothing indicated the list continued past the fifth entry.

`dropUp` is now pinned to `false` on the version picker (`frontend/projects/valtimo/process-link/src/lib/components/configure-building-block-plugins/configure-building-block-plugins.component.html:32`) and the menu has an always-drawn scrollbar (`.version-select ::ng-deep .cds--list-box__menu` in the accompanying `.scss`). The visible height is still Carbon's own cap of 5.5 items, so you see about five at a time and scroll for the rest — but every version is loaded, via `building-block-state.service.ts:131` calling `getAllVersionsForBuildingBlock`, which sets `all=true` and makes the backend return the unpaged list (`BuildingBlockManagementResource.getBuildingBlockDefinitionVersions`).

I could not check this against a running application, so this is from the code as merged rather than from a reproduction.

PR #970, opened 2026-08-27
Show process settings read-only for a finalised case version

Klaas-Ritense:
https://github.com/generiekzaakafhandelcomponent/gzac-issues/issues/506

## Bug

Opening a process of a case version that has been made final showed an almost empty side
panel: clicking a task gave you the process link and nothing else. The marker toggle showed
nothing either, and the diagram could still be changed even though nothing could be saved.

## Fix

The panel now shows the same settings as an editable case version - name, implementation,
listeners, input/output mappings and the rest - all read-only. Markers show again, process
links can be opened to view how they are configured, and nothing on the page can be changed.

Assumed: a process that is read-only on a case version that is not final keeps its editable
process links, since only a final version locks everything.


frontend/projects/valtimo/process-link/src/lib/directives/read-only-content.directive.ts
```diff
@@ -0,0 +1,114 @@
...
+ * Makes the content of the host read-only, and keeps doing so while that content changes. Used for
+ * the process link modal, whose configuration step is rendered by components that are contributed
+ * by plugins and cannot be asked for a read-only mode of their own.
+ *
+ * Two things happen, and both are needed. Marking the individual controls as readonly or disabled
+ * is what makes them *look* unavailable and what a screen reader announces. Taking the host out of
+ * reach of the mouse and the tab order is what actually *guarantees* it: composite widgets such as
+ * the Carbon combo box open their menu from a plain div, which has no disabled state to set, and
+ * a plugin can contribute any widget at all.
+ */
```
sofiaIvarsRitense: I think this comment could be shorter

frontend/projects/valtimo/process-link/src/lib/directives/read-only-content.directive.ts
```diff
@@ -0,0 +1,114 @@
...
+    if (!this._readOnly) {
+      this.setContentPointerEvents('');
+      return;
+    }
+
+    // Nothing inside can be clicked, which covers both the widgets that have no disabled state of
+    // their own and the ones whose disabled state Angular rebinds on every change detection run.
+    // Applied to the children rather than to the host, so that the host itself keeps receiving the
+    // wheel events that scroll a long configuration. Deliberately not `inert`, which would also
+    // hide the configuration from screen readers.
```
sofiaIvarsRitense: I think this comment could be shorter

frontend/projects/valtimo/process-management/src/lib/components/process-management-builder/process-management-builder.component.ts
```diff
@@ -173,6 +178,16 @@ export class ProcessManagementBuilderComponent implements AfterViewInit, OnDestr
   private _activityMarkerUpdateTimeout: any = null;
   private _autofilledElements: AutofilledElement[] = [];
   private _autofilledElementIds: string[] = [];
+  private _viewerActive = false;
+
+  /**
+   * The editor that is on screen. Both are kept loaded with the same diagram, so anything drawn on
+   * top of it - markers, validation highlights, autofill badges - has to go to this one, or it ends
+   * up on the hidden editor. Mirrors the condition the template switches on.
+   */
```
sofiaIvarsRitense: I think this comment is not necessary

sofiaIvarsRitense:
Some inputs like the toggle button look disabled but not read-only
Screenshot 2026-08-28 at 12 40 58
Screenshot 2026-08-28 at 12 40 32

sofiaIvarsRitense (approved):

Klaas-Ritense:
AI: automatic reply, not reviewed by a human.

You're right, and that is not intentional. The toggle in the modal has no `disabled` binding — `frontend/projects/valtimo/process-link/src/lib/components/process-link-modal/process-link-modal.component.html:101` — so its read-only appearance is supposed to come entirely from `ReadOnlyContentDirective`, which adds Carbon's own `cds--toggle--readonly` modifier to `.cds--toggle` (`frontend/projects/valtimo/process-link/src/lib/directives/read-only-content.directive.ts:38`). In your screenshot the track is #8D8D8D with a white knob, which is Carbon's plain *unchecked* toggle; the read-only rule in `@carbon/styles` 1.80.0 makes the track transparent with a 1px border and a dark knob, so the modifier is not taking effect on the live component. The directive's spec only exercises a static `<div class="cds--toggle">` (`read-only-content.directive.spec.ts:78`), never a real `<cds-toggle>` where that class comes from an Angular `HostBinding` — that is the first place I would look, though I could not run the UI here to confirm the exact mechanism.

PR #973, opened 2026-08-27
Create a new draft of a case with building blocks again

Klaas-Ritense:
https://github.com/generiekzaakafhandelcomponent/atlas-internal/issues/689

## Bug

Creating a new draft version of a case that contains building blocks failed with a server
error, and no draft was created.


backend/case/src/main/kotlin/com/ritense/case/service/CaseDefinitionService.kt
```diff
@@ -121,6 +122,16 @@ class CaseDefinitionService(
             require(basedOnCaseDefinition.final) {
                 "Failed to create case-definition-draft. Case-definition with id: '$basedOnCaseDefinitionId' is not final."
             }
+            // A draft copies the configuration of the version it is based on. Copying a broken
```
marijnritense: Not sure I fully get it. If the original has a broken configuration, what if it was put into a broken state because of some database changes? In that case, they'd want to make a draft to fix the issues. I think if copying something fails, it should just continue with everything it can copy.

Klaas-Ritense: Ok, Ive changed that

marijnritense (approved):

PR #1002, opened 2026-09-03
Fix: no error logged when a process ends while a user task is open

Klaas-Ritense:
https://github.com/generiekzaakafhandelcomponent/gzac-issues/issues/859
Also solves: https://github.com/generiekzaakafhandelcomponent/gzac-issues/issues/838

## Bug

When a process was completed or cancelled by a message while a user task on it was still open, the
application log filled with an error and a full stack trace — even though the process had ended exactly
as intended.

## Fix

A process ending while a task is still open no longer logs an error. Anything that genuinely goes wrong
while sending the task update is still reported as before.

### DIFF
--- a/backend/process-document/src/main/kotlin/com/ritense/processdocument/sse/autoconfigure/SseAutoConfiguration.kt
+++ b/backend/process-document/src/main/kotlin/com/ritense/processdocument/sse/autoconfigure/SseAutoConfiguration.kt
@@ -20,6 +20,7 @@ import com.ritense.document.service.DocumentService
 import com.ritense.processdocument.service.ProcessDocumentService
 import com.ritense.processdocument.sse.domain.listener.TaskUpdateListener
 import com.ritense.valtimo.contract.document.CaseDocumentResolver
+import com.ritense.valtimo.service.OperatonProcessService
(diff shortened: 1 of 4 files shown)


ivo-ritense (approved):

ivo-ritense:
I do still see the stacktraces in the log, but it's on debug level so i suppose that makes sense.

Klaas-Ritense:
AI: this reply was written and posted automatically, without human review.

Yes, that is intentional — the stack traces are the same exceptions as before, they just no longer escape the listener as an error. `TaskUpdateListener.resolveAndNotify` now catches them and, once it has confirmed the process instance really has ended, logs them with `logger.debug(e)`: `backend/process-document/src/main/kotlin/com/ritense/processdocument/sse/domain/listener/TaskUpdateListener.kt:81`, and line 67 for the case where the task itself is already gone. The exception is deliberately kept on that DEBUG line, because skipping the notification is only correct when the process has genuinely ended, and the trace is the only way to check that afterwards. At the default `logback.loglevel` of INFO (`backend/apps/gzac/src/main/resources/logback-spring.xml:69`) none of it is printed.

PR #1004, opened 2026-09-03
Fix: deleting a case whose zaak is already gone from the Zaken API

Klaas-Ritense:
https://github.com/generiekzaakafhandelcomponent/gzac-issues/issues/742

## Bug

Deleting a case failed with an error saying the zaak does not exist, when that zaak had
already been removed in the Zaken API. The case stayed behind in GZAC and could never be
deleted.

## Fix

The case is now deleted, and a warning is logged saying the zaak was already gone. When the
Zaken API refuses the deletion for another reason, the case is still kept.

Assumed the ticket's second scenario should keep blocking the deletion rather than become a
configurable option.

### DIFF
--- a/backend/zgw/zaken-api/src/main/kotlin/com/ritense/zakenapi/listener/ZakenApiDocumentDeletedEventListener.kt
+++ b/backend/zgw/zaken-api/src/main/kotlin/com/ritense/zakenapi/listener/ZakenApiDocumentDeletedEventListener.kt
@@ -26,7 +26,9 @@ import com.ritense.zakenapi.link.ZaakInstanceLinkService
 import com.ritense.zakenapi.service.ZaakDocumentService
 import io.github.oshai.kotlinlogging.KotlinLogging
 import org.springframework.context.event.EventListener
+import org.springframework.http.HttpStatus
 import org.springframework.transaction.annotation.Transactional
+import org.springframework.web.client.HttpClientErrorException
 
 @AllOpen
 class ZakenApiDocumentDeletedEventListener(
@@ -55,7 +57,18 @@ class ZakenApiDocumentDeletedEventListener(
                     ZakenApiPlugin::class.java,
                     ZakenApiPlugin.findConfigurationByUrl(link.zaakInstanceUrl)
                 )
-                plugin?.deleteZaak(link.zaakInstanceUrl)
+                try {
+                    plugin?.deleteZaak(link.zaakInstanceUrl)
+                } catch (e: HttpClientErrorException) {
+                    if (e.statusCode == HttpStatus.NOT_FOUND) {
+                        logger.warn(e) {
+                            "Skipping deletion of zaak '${link.zaakInstanceUrl}' of case " +
+                                "'${event.caseDocumentId}': it was not found in the Zaken API"
+                        }
+                    } else {
+                        throw e
+                    }
+                }
             }
 
         }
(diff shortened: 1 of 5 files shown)


backend/zgw/zaken-api/src/main/kotlin/com/ritense/zakenapi/listener/ZakenApiDocumentDeletedEventListener.kt
```diff
@@ -55,7 +57,18 @@ class ZakenApiDocumentDeletedEventListener(
                     ZakenApiPlugin::class.java,
                     ZakenApiPlugin.findConfigurationByUrl(link.zaakInstanceUrl)
                 )
-                plugin?.deleteZaak(link.zaakInstanceUrl)
+                try {
+                    plugin?.deleteZaak(link.zaakInstanceUrl)
```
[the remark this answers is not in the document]
Klaas-Ritense: AI: fixed in c78aa1aac — `deleteRelatedInformatieObjecten` now catches `HttpClientErrorException` around the zaak-scoped `getZaakInformatieObjecten` call, treats a 404 as nothing to clean up (warn and return), and still rethrows other client errors. Regression tests added in `ZaakDocumentServiceTest` for the 404 and for a 403.

Two notes on the shape of the fix. The catch sits inside `ZaakDocumentService` rather than in the listener because that class is `@Transactional`: an exception crossing its proxy would have marked the transaction rollback-only, so catching it in the caller would no longer have saved the case deletion — the same reasoning as the existing comment on the Documenten API delete a few lines below. And the 404 itself is defensive: `zaakinformatieobjecten?zaak=<url>` is a filtered list endpoint, so a stale filter plausibly yields an empty list or a 400 rather than a 404 depending on the implementation. The linked issue (gzac-issues#742) flags the exact status code as unverified against a live OpenZaak, so this branch is robustness rather than a reproduced failure.

backend/zgw/zaken-api/src/main/kotlin/com/ritense/zakenapi/service/ZaakDocumentService.kt
```diff
@@ -155,7 +155,24 @@ class ZaakDocumentService(
...
         ) { "Could not find ${ZakenApiPlugin::class.simpleName} configuration for zaak with url: $zaakInstanceUrl" }
 
-        zakenApiPlugin.getZaakInformatieObjecten(caseDocumentId, zaakInstanceUrl).forEach { zaakInformatieobject ->
+        val zaakInformatieObjecten = try {
+            zakenApiPlugin.getZaakInformatieObjecten(caseDocumentId, zaakInstanceUrl)
+        } catch (e: HttpClientErrorException) {
+            if (e.statusCode == HttpStatus.NOT_FOUND) {
+                // The catch belongs here rather than in the caller: this class is @Transactional,
+                // so an exception crossing its proxy would have marked the transaction
+                // rollback-only and the case deletion could no longer be saved.
```
Klaas-Ritense: Remove this comment

Klaas-Ritense: AI: fixed in 9680ab746 — removed the three-line comment above the `logger.warn` in `deleteRelatedInformatieObjecten` that explained why the catch sits in `ZaakDocumentService` rather than in the listener. `:backend:zgw:zaken-api:test` still green.

ivo-ritense (approved):

ivo-ritense (approved):

PR #1011, opened 2026-09-04
Fix: currency field default value shown a hundred times too small

Klaas-Ritense:
https://github.com/generiekzaakafhandelcomponent/gzac-issues/issues/176

## Bug

A currency field with a default value showed a hundredth of the amount — a field set to 100 appeared
as € 1,00. Opening the form also saved that smaller amount back onto the case.

## Fix

A currency field now shows its default value in full, and opening the form leaves the stored amount
unchanged.

Not verified against a running app: the dev stack could not be started, so this is covered by unit tests only.


sofiaIvarsRitense (approved):

sofiaIvarsRitense:
Testing finding:
- The default currency value is correctly displayed but the original form doesn't keep the correct value.

https://github.com/user-attachments/assets/5cc64dcf-243f-4f88-9893-080767d8c187

Klaas-Ritense:
AI: not changed — I could not reproduce the wrong write-back from the component alone, so I have not guessed at a fix.

What I checked against `@tadashi/currency` 3.4.0: the display path now masks with a fresh options object (`viaInput: false`), which is the half you confirmed works. The write-back path goes through `getUnmasked()` — and that method takes **no** arguments. It always unmasks `this.input.value`, the DOM value, and silently ignores the `currencyForm` value the component hands it. In a unit test that still round-trips (`€ 100,00` → `100`); the six new specs and the rest of `@valtimo/components` are green (74/74) after syncing with `next-minor`.

So whatever drops the amount is not visible from the component in isolation. My best hypothesis is the ordering in `create-custom-component.ts`, where `attach()` binds `customOptions` before it calls `restoreValue()` — that lets `ngOnChanges` fire, and emit, while `value` is still null.

What would decide it: which value the case ends up holding (`0`, `1`, or the previous amount), and whether it changes when the form opens or only on submit. Either that detail, or a running dev stack to reproduce against, and I can carry on. I did not want to change the emit path on a guess, because it would move the stored value of every currency field, not just ones with a default.

Klaas-Ritense:
Test finding:
I found the same issue as @sofiaIvarsRitense:
1. Create a new Form.io
2. Add a currency component
3. Set default value to €100,00
4. Save the currency component, but dont save the formio
5. The formio builder does not show the new default 
6. Open the recently added currency component
7. Navigate to the Default value field
8. Find it does NOT show €100,00
9. Save the formio anyway
10. Open any task that uses the formio
11. Find that it DOES show the €100,00 default value

Klaas-Ritense:
AI: not changed — I found the mechanism behind steps 6–8, but the fix is a builder-behaviour decision rather than a currency bug, so I have not made it on a guess.

What I read in `formiojs` 4.19.5 (the pinned version):

- The edit dialog's **Default Value** field is declared as a plain `textfield` in the base data tab (`components/_classes/component/editForm/Component.edit.data.js:19-25`). It is not the edited component's own type.
- `WebformBuilder.updateComponent()` is what swaps it for the real component type, and the swap is gated on `changed && !isNil(changed.value)` (`WebformBuilder.js:1237`). Opening the dialog calls `updateComponent(componentCopy)` with no `changed` argument, and the `editForm.on('change')` handler that would supply one is registered *after* that call.

So on reopening, the Default Value field is still the TextField instance built at attach time, and it shows the raw stored number `100` instead of `€ 100,00`. Change any other setting in the dialog and it rebuilds as a currency field and formats correctly. That also fits step 11: the schema holds `defaultValue: 100`, which is why the task form is right — nothing is lost, only the builder's own rendering of it.

The candidate fix is for `currencyEditForm()` to replace the data tab's `defaultValue` entry with a `currency`-typed one, so the instance is the right class from the start. I did not apply it because the `customOptions` (locale, currency, allow-empty) would then be pinned to the schema defaults until the first change event rebuilds the field with the values actually chosen in the same dialog — that is a behaviour choice, and not one I can observe passing from a unit test, since it only appears in a running builder.

What would decide it: whether the default-value field showing `nl-NL`/`EUR` formatting until the dialog's first change is acceptable, or whether it has to mirror the chosen locale immediately.

On step 5 (the canvas not showing the new default) I have no answer yet — the canvas path does reach `restoreValue()` → `setDefaultValue()`, so on reading it should render, and pinning down why it does not needs a running builder.

This pass also synced the branch with `next-minor` — one conflict, where both sides had appended a row to the 13.45.0 Bugfixes table, and I kept both in the table's existing alphabetical order — and re-ran `@valtimo/components`: 74/74 green.

PR #1012, opened 2026-09-06
Fix: repeated clicks on Save upload the same document more than once

Klaas-Ritense:
https://github.com/generiekzaakafhandelcomponent/gzac-issues/issues/865

## Bug

On a case's Documents tab, Save in the file metadata window stayed active while the upload was
still running. Clicking it again filed the same document again, so the case ended up with one copy
per click.

## Fix

Save is now switched off as soon as it is clicked and stays off until the upload finishes, so a
document is added to the case once however many times Save is clicked.


sofiaIvarsRitense (approved):

sofiaIvarsRitense:
Testing finding:
- I would disable the 'Cancel' button as well, otherwise the following situation can be confusing:

https://github.com/user-attachments/assets/4684c272-73ee-4c6d-a96f-e467ca23110c

Klaas-Ritense:
AI: fixed in 3428a7b08 — Cancel is now switched off while an upload is in flight, the same way Save already was (`[disabled]="obs.disabled || uploading"` on the metadata window footer). The window therefore cannot be dismissed mid-upload, so the case worker is not left unsure whether the document was filed. The guard sits on the button only, not in `closeModal()`, because `save()` calls that method itself to close the window when every field is hidden — guarding it there would wedge the window open on that path. `npm run libs:test:zgw` passes 2/2 and `ng lint @valtimo/zgw` is clean.

PR #1013, opened 2026-09-07
Fix: reopening a plugin action showed a different plugin's action

Klaas-Ritense:
https://github.com/generiekzaakafhandelcomponent/gzac-issues/issues/386

## Bug

When you reopened an already configured plugin action on a service task in Proceskoppelingen, a
different plugin's action could be shown if two plugins happen to use the same action name.

## Fix

Reopening a configured plugin action now shows the action of the plugin it was actually set up
with, with its saved settings.

Assumption: when the plugin configuration a process link points at no longer exists, the step now
reports that no configuration is available instead of showing a best guess.


frontend/projects/valtimo/process-link/src/lib/services/plugin-state.service.ts
```diff
@@ -168,18 +151,32 @@ export class PluginStateService {
...
-    if (processLink.pluginDefinitionKey) {
+    if (processLink?.pluginDefinitionKey) {
       return of(processLink.pluginDefinitionKey);
     }
 
-    // Otherwise, derive it from plugin specifications using the action key
+    // An action key can occur in several plugins, so the configuration the link points at decides which one
+    if (processLink?.pluginConfigurationId) {
+      return this.pluginManagementService
+        .getAllPluginConfigurations()
```
Klaas-Ritense: Please add a new endpoint to get the plugin configuration by UUID

Klaas-Ritense: AI: fixed in 972b14af0 — added `GET /api/v1/plugin/configuration/{pluginConfigurationId}` (ADMIN, like the other configuration endpoints) backed by a new `PluginService.findPluginConfiguration(id)`, which answers 404 rather than 500 when the configuration is gone. `PluginStateService` now resolves the process link's configuration through it instead of fetching every configuration and matching client-side. Resource tests cover the found and not-found cases. No release note added for the endpoint itself: it only serves the bugfix that is already noted.

frontend/projects/valtimo/process-link/src/lib/services/plugin-state.service.ts
```diff
@@ -168,18 +151,32 @@ export class PluginStateService {
...
+        .pipe(
+          map(
+            configurations =>
+              configurations.find(
+                configuration => configuration.id === processLink.pluginConfigurationId
+              )?.pluginDefinition?.key
```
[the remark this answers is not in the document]
Klaas-Ritense: AI: fixed in 972b14af0 — `selectProcessLink` now clears the selected definition, configuration and action before loading the new link, and the by-id lookup catches the 404 a deleted configuration answers. Reopening a link whose `pluginConfigurationId` no longer exists therefore shows nothing rather than the previously opened plugin. Two regression tests cover it: a valid configuration followed by an unknown id clears all three, and the definition key resolves to undefined instead of falling through to the ambiguous action key.

frontend/projects/valtimo/process-link/src/lib/services/plugin-state.service.ts
```diff
@@ -114,6 +97,11 @@ export class PluginStateService {
   }
 
   selectProcessLink(processLink: ProcessLink): void {
+    // The previous link's plugin, configuration and action must not survive into this one. A link
+    // whose configuration has since been deleted resolves to nothing, and without clearing first
+    // the plugin that was on screen before would stay there as though it were this link's own
```
Klaas-Ritense: Remove code comment

Klaas-Ritense: AI: fixed in 0e8fc7e58 — removed the three-line block comment above `this.clearPluginSelection()` in `selectProcessLink`. The clearing call and its two regression tests are unchanged; only the comment is gone. The other comments this PR added elsewhere in the file are left as they are, since this thread pointed at that block.

frontend/projects/valtimo/process-link/src/lib/services/plugin-state.service.ts
```diff
@@ -114,6 +97,8 @@ export class PluginStateService {
   }
 
   selectProcessLink(processLink: ProcessLink): void {
+    this.clearPluginSelection();
```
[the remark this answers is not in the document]
Klaas-Ritense: AI: fixed in 2f4f153e7 — `selectProcessLink` now emits on a `_cancelPluginLoad$` subject before it clears the selection, and all three lookups (`getPluginDefinitionKeyForProcessLink`, the `getPluginDefinitions` follow-up, and the configuration fetch) are piped through `takeUntil(this._cancelPluginLoad$)`. A lookup still in flight for the previous link is therefore unsubscribed rather than left to write over the new selection. Regression test `ignores a lookup that only finishes after another link was selected` holds link A on a `Subject`, selects link B, then lets A emit; against the previous code it failed with `Expected 'xential' to be 'smart-documents'`, which is this PR's own bug arriving by the slow path. I used `takeUntil` rather than restructuring onto `switchMap`, since the loads are three separate subscriptions driven by an imperative call and cancellation was the part that was missing.

frontend/projects/valtimo/process-link/src/lib/services/plugin-state.service.ts
```diff
@@ -168,24 +155,41 @@ export class PluginStateService {
...
   }
 
+  // A configuration the link still records but that no longer exists answers 404, which is not an
+  // error worth surfacing here — the link simply has no plugin to show
+  private getPluginConfigurationForProcessLink(
+    processLink: ProcessLink
+  ): Observable<PluginConfiguration | undefined> {
+    return this.pluginManagementService
+      .getPluginConfiguration(processLink.pluginConfigurationId)
+      .pipe(catchError(() => of(undefined)));
```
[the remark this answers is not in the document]
Klaas-Ritense: AI: fixed in 2f4f153e7 — `getPluginConfigurationForProcessLink` now maps only a 404 to `undefined` and rethrows anything else, so a 500 no longer reads as "this link has no configuration". Covered by `surfaces an error that is not a missing configuration`, which asserts the 500 reaches a `pluginDefinitionKey$` subscriber. Two notes on the shape: the check is `error?.status === 404` rather than an `instanceof HttpErrorResponse` guard, so it holds for the plain error objects the spec throws as well as real `HttpErrorResponse`s; and the eager loads inside `loadPluginDefinitionForProcessLink` got an `error` handler that clears the selection, because they are fire-and-forget with nothing to report to — the rethrown error surfaces through the public `pluginDefinitionKey$` stream, which is where a component can act on it.

sofiaIvarsRitense (approved):

